PR #166, opened 2016-11-30
Feature/gui editor macros

### LINKED ISSUE #146: Macros
Macros in Editor

FelixOpolka:
closes #146


goldsborough:
This is a lot of code 😮

CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/source)
 # TESTS
 ########################################
 
-option(ERA_GP_SIM_TESTS "Enable tests" OFF)
+option(ERA_GP_SIM_TESTS "Enable tests" ON)
```
goldsborough: Eh, no. Unless you remember to turn this off in a few months :blush: The idea is that users who want to build our project without tests don't have to compile the tests. Either do this locally but don't push, or just pass `DERA_GP_SIM_TESTS=ON`. See [here](http://stackoverflow.com/questions/18435516/how-to-set-a-cmake-option-at-command-line)

include/arch/common/abstract-instruction-node.hpp
```diff
@@ -39,8 +40,19 @@ class AbstractInstructionNode : public AbstractSyntaxTreeNode
...
+     * Note that this member function should be independent from any children as it may be called for the user
+     * to display information how to define a valid instruction of this kind.
+     * \return A Translateable text containing information dedicated to the user about this instruction
+     */
+    virtual const Translateable& getInstructionDocumentation() const = 0;
+
 protected:
 
+    /**
+     * Returns the InstructionInformation of this Instruction
```
goldsborough: Prefer to write `\return the InstructionInformation ...` Because then it gets rendered better by doxygen

include/arch/common/abstract-register-node-factory.hpp
```diff
@@ -30,6 +31,8 @@ class AbstractRegisterNodeFactory {
  public:
   using Node = std::unique_ptr<AbstractSyntaxTreeNode>;
 
+  explicit AbstractRegisterNodeFactory(const Architecture& arch) {}
```
goldsborough: What's the point?

include/arch/common/node-factory-collection.hpp
```diff
@@ -41,6 +42,7 @@ class AbstractRegisterNodeFactory;
 class NodeFactoryCollection {
  public:
   using Node = std::unique_ptr<AbstractSyntaxTreeNode>;
+    using InstrNode = std::unique_ptr<AbstractInstructionNode>;
```
goldsborough, line 45: No abbreviations, `InstructionNodePointer`

goldsborough:
This is too much code 😢

goldsborough (commented):
I don't really know what here is new. I already reviewed some of this i other branches?

include/parser/macro-information.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+#ifndef ERAGPSIM_PARSER_MACRO_INFORMATION_HPP
+#define ERAGPSIM_PARSER_MACRO_INFORMATION_HPP
+
+#include "parser/code-position.hpp"
+
+/**
+ * Class containing macro information.
+ */
+class MacroInformation {
```
goldsborough, line 29: Did you writ this class? If so put the implementations into the implementation file

include/ui/clipboard-adapter.hpp
```diff
@@ -0,0 +1,46 @@
...
+  Q_OBJECT
+ public:
+  explicit ClipboardAdapter(QObject *parent = 0);
+
+  /**
+  \brief setText Allows QML instances to set the clipboard
+  data.
+  \param text New clipboard text.
+  */
+  Q_INVOKABLE void setText(QString text);
```
goldsborough: `const&`

goldsborough: Please be careful about this

source/parser/intermediate-instruction.cpp
```diff
@@ -188,6 +189,31 @@ void IntermediateInstruction::insertIntoArguments(const std::string& name,
   replaceInVector(_targets, name, value);
 }
 
+std::string IntermediateInstruction::toString() const {
+    std::string str = _name;
+
+    // Append targets
+    for (size_t i = 0; i < _targets.size(); i++) {
+      if (i != 0)
```
goldsborough: I guess you didn't write this, but either way there should be braces around these `if`/`else`s

source/ui/Components/editor/MacroTriangleButton.qml
```diff
@@ -0,0 +1,55 @@
...
+            text: control.text
+            font.pointSize: 10
+            color: "gray"
+        }
+        background: Rectangle {
+            visible: false
+        }
+    }
+    onClicked: {
+        text = (text == "\u25b6") ? "\u25bc" : "\u25b6";
```
goldsborough: Please add comments for what these unicode chars represent

source/ui/Components/editor/MacroTriangleButton.qml
```diff
@@ -0,0 +1,55 @@
...
+        onExpandedChanged(macroIndex);
+    }
+
+    function setExpanded() {
+        text = "\u25bc";
```
goldsborough: same (or make them constants/variables with descriptive names)

goldsborough (approved):

source/ui/Components/editor/Editor.qml
```diff
@@ -170,102 +199,488 @@ ScrollView {
...
+                    // Create display objects.
+                    for (var macroIndex = 0; macroIndex < macros.length; ++macroIndex) {
+                        var macro = macros[macroIndex];
+                        // Find the line for the macro expansion.
+                        var linePosition = TextUtilities.getPositionOfOccurence(text, "\n", macro["startLine"]+1);
+                        // Create display objects (i.e. sub-editor and triangle)
+                        var macroDisplayObject = {};
+                        // Add sub-editor to display object
+                        var yPos = textArea.positionToRectangle(linePosition).y + textArea.cursorRectangle.height
+                        var subeditorComponent = Qt.createComponent("MacroSubeditor.qml");
```
janschopohl: why don't you create the Component once like this:
```
Component {
id: subeditorComponent
    MacroSubeditor {
    }
}
```
You wouldn't have to load the component dynamically every time you call this method, if you do it like that.

source/ui/Components/editor/Editor.qml
```diff
@@ -170,102 +199,488 @@ ScrollView {
...
+                        // Delete old line numbers
+                        for (var i = 0; i < _lineNumberObjects.length; ++i) {
+                            _lineNumberObjects[i].destroy();
+                        }
+                        _lineNumberObjects = [];
+
+                        // Add new line numbers.
+                        var lineNumberLabels = getLineNumberLabels();
+                        for (var line = 0; line < textArea.lineCount; ++line) {
+                            var lineNumber = Qt.createQmlObject('import QtQuick 2.6; Text {color: "gray"; font: textArea.font}',
```
janschopohl: pretty much the same as the `createComponent` call above:
If you don't have to dynamically create a qml component/object, you can already define the component in the qml code. And `createQmlObject` should be used, if you want to dynamically change the kind of object you create at runtime from a string and can't define it in the code. I would prefer the method with one Component and `createObject` calls.

source/ui/Components/editor/Editor.qml
```diff
@@ -170,102 +199,488 @@ ScrollView {
...
+                                                                lineNumbersBar);
+                            // TODO: Find out why we need that correction factor... So strange;
+                            // Not required before but now even required when using Repeater...
```
janschopohl: It is probably quite important to find out the reason for that, as a hardcoded factor like this almost guarantees problems on some maschines.
Edit: Just checked, line numbers don't line up right for me...

source/ui/Components/editor/Editor.qml
```diff
@@ -170,102 +199,488 @@ ScrollView {
...
                     y: textArea.textMargin/2
 
-                    Repeater {
-                        model: textArea.lineCount
-                        delegate: Text {
-                            color: "gray"
-                            font: textArea.font
-                            text: index + 1
-                            height: fontMetrics.height
+                    property var _lineNumberObjects: []
```
janschopohl: I think you could do this way easier:
Take the old implementation and set `text: lineNumberLabels[index]`. Then create a `updateLineNumbers()` signal in qml, which indicates to the text delegates that the lineNumberLabels array changed. In the text delegates, you add a `Connections` elements which updates the text in `onUpdateLineNumbers: text = lineNumberLabels[index]`. You can also try to only do a full update of the line number objects/labels if there really are macros.
That might work and would eliminate the need to recreate all line number objects all the time.

The thing that I'm not really happy about in the current implementation is that you always have to do this potentially expensive line number updating, even if there are no macros and there is no need for this.

source/ui/Components/editor/MacroSubeditor.qml
```diff
@@ -0,0 +1,92 @@
...
+    anchors.left: parent.left
+    anchors.right: parent.right
+
+    property alias text: macroTextEdit.text
+    property var expandedHeight: 0
+    height: 0
+    clip: true
+
+    function expand() {
+        height = expandedHeight;
```
janschopohl, line 34: You could do a nice expanding animation here, but that is up to you.

source/ui/editor-component.cpp
```diff
@@ -47,6 +47,25 @@ EditorComponent::EditorComponent(QQmlContext *projectContext,
         throwRuntimeError(validationResult);
       });
 
+  parserInterface.setSetMacroListCallback([this](
+      const std::vector<MacroInformation> &macroList) {
+    QVariantList updatedMacroList;
+    for (auto it = macroList.begin(); it != macroList.end(); ++it) {
```
janschopohl: you could also write `for (const auto& macroInformation : macroList) {...}`

source/ui/Components/editor/Editor.qml
```diff
@@ -170,102 +199,488 @@ ScrollView {
                     id: fontMetrics
                     font: textArea.font
                 }
+
```
janschopohl, line 201: There is quite much code and logic in the Editor.qml file now, which makes it kind of a pain to go through. We should consider refactoring this file and pulling as many components as possible into separate qml files to simplify the editor, if we have time for that (There are probably more important things right now).

source/ui/Components/editor/MacroSubeditor.qml
```diff
@@ -0,0 +1,92 @@
...
+    TextEdit {
+        id: macroTextEdit
+        anchors.left: macroSidebar.right
+        anchors.right: parent.right
+        anchors.top: parent.top
+        anchors.topMargin: -2
+        anchors.bottom: parent.bottom
+
+        readOnly: true
+        text: "mov eax, 0"
```
janschopohl: Is this still needed?

source/ui/output-component.cpp
```diff
@@ -70,7 +69,8 @@ void OutputComponent::updateMemory(size_t address, size_t length) {
 void OutputComponent::putMemoryValue(int address,
                                      QList<bool> memoryContentBitList) {
   // Round up size of altered content up to bytes.
-  auto memoryValueSize = Utility::roundToBoundary(memoryContentBitList.size(), 8);
+  auto memoryValueSize =
+      Utility::roundToBoundary(memoryContentBitList.size(), 8);
```
goldsborough: This is called `discreteCeiling` now

source/ui/Components/Common/TextUtilities.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Returns a string with the substring beginning at a (inclusive) and ending at b (exclusive) removed from text.
+function remove(string, a, b) {
+    return string.slice(0, a) + string.slice(b, string.length);
+}
+
+// Returns the number of occurences of a given searchString inside a given text.
+function numberOfOccurences(string, searchString) {
+    string += "";
```
goldsborough: why?

source/ui/Components/Common/TextUtilities.js
```diff
@@ -0,0 +1,70 @@
...
+    searchString += "";
```
goldsborough: why?

source/ui/Components/Common/TextUtilities.js
```diff
@@ -0,0 +1,70 @@
...
+    if (searchString.length <= 0) return (string.length + 1);
```
goldsborough: why the parantheses?

source/ui/Components/Common/TextUtilities.js
```diff
@@ -0,0 +1,70 @@
...
+
+    var index = -1;
+    var offset = 0;
+    var count = 0;
+    while ((index = string.indexOf(searchString, offset)) != -1) {
+        count++;
+        offset = (index + searchString.length);
```
goldsborough: don't need the parentheses

source/ui/Components/Common/TextUtilities.js
```diff
@@ -0,0 +1,70 @@
...
+    var lineStart = text.slice(0, position).lastIndexOf("\n") + 1;
+    if (lineStart == -1) { lineStart = 0; }
+    return lineStart;
+}
+
+// Returns the position of the end of the line (i.e. the next character would be the newline character)
+// relative to the entire editor text.
+function getLineEndForPosition(text, position) {
+    var lineEnd = textArea.text.slice(position).indexOf("\n") + position;
+    if (lineEnd == -1) { lineEnd = text.length; }
```
goldsborough: don't need the braces here

source/ui/Components/Common/TextUtilities.js
```diff
@@ -0,0 +1,70 @@
...
+    return lineEnd;
```
goldsborough: You can write this as `return (lineEnd == -1) ? text.length : lineEnd;`

source/parser/intermediate-representator.cpp
```diff
@@ -24,6 +24,21 @@
...
 
+void IntermediateRepresentator::generateMacroInformation(
+    FinalRepresentation& representation) {
+  for (const auto& i : _commandList) {
+    if (dynamic_cast<IntermediateMacroInstruction*>(i.get()) == nullptr)
+      continue;
+
+    MacroInformation info(
+        static_cast<IntermediateMacroInstruction&>(*i).toString(),
+        CodePositionInterval(CodePosition(i->lines().lineStart, 0),
```
goldsborough: Use temporary variables here, not one big inline statement

source/parser/intermediate-representator.cpp
```diff
@@ -24,6 +24,21 @@
 #include "parser/memory-allocator.hpp"
 #include "parser/symbol-table.hpp"
 
+void IntermediateRepresentator::generateMacroInformation(
+    FinalRepresentation& representation) {
+  for (const auto& i : _commandList) {
+    if (dynamic_cast<IntermediateMacroInstruction*>(i.get()) == nullptr)
```
goldsborough: `dynamic_cast`s are very expensive in c++, because the runtime must traverse the entire inheritance tree to make sure these classes are related. Could we do this differently (maybe, in the future, use tags in the classes, i.e. enum members)

source/parser/intermediate-instruction.cpp
```diff
@@ -188,6 +189,31 @@ void IntermediateInstruction::insertIntoArguments(const std::string& name,
...
+        str += ", ";
+      else
+        str += " ";
+      str += _targets[i];
+    }
+
+    // Append sources
+    for (size_t i = 0; i < _sources.size(); i++) {
+      if (i == 0 && _targets.size() == 0)
```
goldsborough: Please use braces here

include/parser/macro-information.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+
+ private:
+  /**
+   * \brief The internal code attribute.
+   */
+  std::string _code;
+
+  /**
+   * \brief The internal position attribute.
```
goldsborough: The first line of our documentation is implicitly the `\brief`, so you can just write this on one line as `/** <documentation> */`

include/parser/macro-information.hpp
```diff
@@ -0,0 +1,67 @@
...
+   * \param code Parsed macro code
+   * \param position Position of the macro call
+   */
+  MacroInformation(const std::string& code,
+                   const CodePositionInterval& position)
+  : _code(code), _position(position) {
+  }
+
+  /**
+   * Returns the expanded code for this macro.
```
goldsborough, line 39: `\return The expanded code for this macro.`

include/parser/macro-information.hpp
```diff
@@ -0,0 +1,67 @@
...
+   */
+  const std::string& macroCode() const {
```
goldsborough: `noexcept`

goldsborough (changes requested):

include/parser/macro-information.hpp
```diff
@@ -32,34 +33,29 @@ class MacroInformation {
...
 
   /**
-   * \brief Returns the position of this macro.
+   * Returns the position of this macro.
    * \return The position of the macro.
    */
-  const CodePositionInterval& position() const {
-    return _position;
-  }
+  const CodePositionInterval& position() const;
```
goldsborough, line 48: In the future, you can make this kind of getter `noexcept`.

PR #170, opened 2016-12-01
Hotfix/memory

### LINKED ISSUE #165: call gui callback on memory deserialization
The memory should call the gui update callback for the whole memory when a snapshot is loaded. Currently loading a snapshot won't set non zero values to zero.

ProfPille:
fix #165

### DIFF
--- a/include/core/memory.hpp
+++ b/include/core/memory.hpp
@@ -110,18 +110,40 @@ class Memory {
    * \param amount Number of Bytes comprising the value
    * \returns MemoryValue holding the data stored in the Memory at
    *          [address;address+amount[
+   * \throws AssertionError iff [address; address+amount[ is not subset of
+   *         [0; _byteCount[
    */
-  MemoryValue
-  get(const std::size_t address, const std::size_t amount = 1) const;
+  MemoryValue get(std::size_t address, std::size_t amount = 1) const;
+  /**
+   * \brief Returns a MemoryValue holding the data stored in the Memory at
+   *        [address; address+amount[
+   * \param address Starting location of the Value in the Memory
+   * \param amount Number of Bytes comprising the value
+   * \returns MemoryValue holding the data stored in the Memory at
+   *          [address;address+amount[ with 0 for all bytes not in
+   *          [0; _byteCount[
+   */
+  MemoryValue tryGet(std::size_t address, std::size_t amount = 1) const;
+  /**
+   * \brief Writes value into the Memory at address
+   * \param address Starting address of the to be overwritten value
+   * \param value Value to write
+   * \param ignoreProtection if this is true do ignore all protection
+   * \throws AssertionError iff [address; address+(value.size()/_byteSize)[ is
+   *         not subset of [0; _byteCount[ or value.size()%_byteSize != 0
+   */
+  void put(std::size_t address,
+           const MemoryValue &value,
+           bool ignoreProtection = false);
   /**
    * \brief Writes value into the Memory at address
    * \param address Starting address of the to be overwritten value
    * \param value Value to write
    * \param ignoreProtection if this is true do ignore all protection
    */
-  void put(const std::size_t address,
-           const MemoryValue &value,
-           bool ignoreProtection = false);
+  void tryPut(std::size_t address,
+              const MemoryValue &value,
+              bool ignoreProtection = false);
   /**
    * \brief Writes value into the Memory at address and returns the previous
    *        value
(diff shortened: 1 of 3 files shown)


include/core/memory.hpp
```diff
@@ -110,29 +111,65 @@ class Memory {
...
-  MemoryValue
-  get(const std::size_t address, const std::size_t amount = 1) const;
+  MemoryValue get(std::size_t address, std::size_t amount = 1) const;
+  /**
+   * \brief Returns a MemoryValue holding the data stored in the Memory at
+   *        [address; address+amount[
+   * \param address Starting location of the Value in the Memory
+   * \param amount Number of Bytes comprising the value
+   * \returns MemoryValue holding the data stored in the Memory at
+   *          [address;address+amount[ with 0 for all bytes not in
```
goldsborough, line 127: Maybe use markup here, with backticks: \`[address;address+amount\`

include/core/memory.hpp
```diff
@@ -110,29 +111,65 @@ class Memory {
...
    *        value
    * \param address Starting address of the to be overwritten value
    * \param value Value to write
    * \param ignoreProtection if this is true do ignore all protection
    * \returns Value that was overwritten
+   * \throws AssertionError iff [address; address+(value.size()/_byteSize)[ is
+   *         not subset of [0; _byteCount[ or value.size()%_byteSize != 0
    */
-  MemoryValue set(const std::size_t address,
+  MemoryValue set(std::size_t address,
```
goldsborough, line 161: If you use `using size_t = std::size_t` at the top of the class, you can just use `size_t`

tests/core/memory-test.cpp
```diff
@@ -67,7 +67,7 @@ TEST(memory, readWrite) {
 }
 
 TEST(memory, serialization) {
-  constexpr std::size_t memorySize = 1024 * 64;
+  constexpr std::size_t memorySize = 1 * 64 - 0;
```
goldsborough: ?

include/core/memory.hpp
```diff
@@ -235,9 +272,7 @@ class Memory {
   /**
    * \brief This function gets called for every changed area in Memory
    */
-  std::function<void(const std::size_t, const std::size_t)> _callback = [](
-      const std::size_t, const std::size_t) {};
-  /**< Brief This function gets called for every changed area in Memory*/
+  CallbackFunction _callback = [](const std::size_t, const std::size_t) {};
```
goldsborough: Do not define variable in the header! That is *very* un-C++-y! Use the constructor's initializer list

include/core/memory.hpp
```diff
@@ -247,12 +282,52 @@ class Memory {
...
+   *        notifies the Gui of the change
+   */
+  void _wasUpdated();
+  /**
    * \brief This Method is called whenever something in the Memory changes and
    *        notifies the Gui of the change
    * \param address address of the updated value
    * \param amount length of the updated value
    */
-  void _wasUpdated(const std::size_t address, const std::size_t amount = 1);
```
goldsborough, line 255: The ordering of things in this file is all off. It should be:

1. Typedefs and nested type declarations
2. Static variables
3. Static functions
4. Private member functions
5. Private variables

source/core/memory.cpp
```diff
@@ -57,69 +57,99 @@ std::size_t Memory::getByteCount() const {
...
-  }
-  // if no characters have been printed print 0
-  if (zero) {
-    strm.put('0');
-  }
+  std::string s = value.toHexString(false, true);
+  strm << s;
 }
 
 std::pair<std::map<std::string, std::size_t>,
```
goldsborough: I bet I can write `using MapPair = ...`  or whatever faster than you can write this type. This seems literally why `typedefs`/`using` exists 😄

include/core/memory.hpp
```diff
@@ -247,12 +282,52 @@ class Memory {
...
   /**
+   * \brief Returns a MemoryValue holding the data stored in the Memory at
+   *        [address; address+amount[, undefined behaviour if
+   *        [address; address+amount[ is not subset of [0;_byteCount[
+   * \param address Starting location of the Value in the Memory
+   * \param amount Number of Bytes comprising the value
+   * \returns MemoryValue holding the data stored in the Memory at
+   *          [address;address+amount[
+   */
+  MemoryValue _get(std::size_t address, std::size_t amount = 1) const;
```
goldsborough, line 294: Use clang format.

source/core/memory.cpp
```diff
@@ -57,69 +57,99 @@ std::size_t Memory::getByteCount() const {
...
+  std::size_t amount{value.getSize() / _byteSize};
   MemoryValue prev{get(address, amount)};
   put(address, value, ignoreProtection);
   return prev;
 }
 
+MemoryValue Memory::trySet(std::size_t address,
+                           const MemoryValue& value,
+                           bool ignoreProtection) {
+  std::size_t amount{value.getSize() / _byteSize};
```
goldsborough, line 146: `auto amount = value.getSize() / byteSize`

source/core/memory.cpp
```diff
@@ -57,69 +57,99 @@ std::size_t Memory::getByteCount() const {
...
+  MemoryValue prev{tryGet(address, amount)};
```
goldsborough, line 147: s/prev/previous

goldsborough (approved):

PR #176, opened 2016-12-07
Feature/gui register component

### LINKED ISSUE #156: Improved register model

### LINKED ISSUE #172: Set all registers' format at once

FelixOpolka:
closes #156 closes #172 
* The register model now offers a variety of data roles to provide the registers’ content instead of using invokable data-functions.
* DefaultRegister text field now uses custom formatting method instead of input mask
* Added context menu (right click) for formatting all registers at once


source/ui/Components/Register/DefaultRegister.qml
```diff
@@ -39,17 +61,36 @@ TextField {
...
+
+    Keys.onPressed: {
+        if (event.key === Qt.Key_Tab) {
+            event.accepted = true;
+            registerTextField.focus = false;
+        }
     }
+
+    function format(registerContentString) {
+        registerContentString = registerContentString.replace(/ /gi, '');
```
goldsborough: I think you can get rid of the 'i' for the regex.

source/ui/register-model.cpp
```diff
@@ -91,11 +97,41 @@ QVariant RegisterModel::data(const QModelIndex &index, int role) const {
...
+      case RegisterInformation::Type::VECTOR:
+          return "Vector";
+      case RegisterInformation::Type::FLAG:
+          return "Flag";
+      case RegisterInformation::Type::LINK:
+          return "Link";
+      case RegisterInformation::Type::PROGRAM_COUNTER:
+          return "ProgramCounter";
+      }
+  case BinaryDataRole:
```
goldsborough: You could replace this with an appropriate map and then do:
```
auto name = _memoryAccess.getRegisterValue(registerItem->getName()).get()));
return QString::fromStdString(conversions[format](name);
```

source/ui/register-model.cpp
```diff
@@ -181,36 +217,31 @@ int RegisterModel::columnCount(const QModelIndex &parent) const {
...
-              registerContentCleared.toStdString(), registerItem->getSize());
-    } else if (*dataFormat == "Decimal (Signed)") {
-      registerContentMemoryValue =
-          StringConversions::signedDecStringToMemoryValue(
-              registerContentCleared.toStdString(), registerItem->getSize());
-    } else if (*dataFormat == "Flag") {
-      registerContentMemoryValue = StringConversions::binStringToMemoryValue(
-          registerContentCleared.toStdString(), registerItem->getSize());
-    }
+  if (dataFormat == "Binary") {
```
goldsborough: This could also be replaced with an appropriate `static const` mapping

goldsborough (approved):

riedersa:
I have reviewed your code and it Looks fine and works on my PC. Except Changing the global Settings for representation. Should this work?

sabrie28:
I habe reviewed your Code and it Looks fine to me and works on my PC. Except changing the global Settings for Representation. Should this work?

riedersa:
Update: Buttons in Toolbar will probably be removed, so it is fine.

riedersa (approved):

PR #180, opened 2016-12-14
Hotfix/assemble 3

### LINKED ISSUE #135: Fix assemble
Write tests at least for each category of formats to return the correct assembled form.

goldsborough:
closes #135


source/arch/riscv/format.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+MemoryValue assemble(const InstructionInformation& instructionInformation,
+                     const Operands& operands) {
+  const auto assembler = getAssembler(instructionInformation.getFormat());
+  return assembler(instructionInformation.getKey(), operands);
+}
+
+MemoryValue R(const InstructionKey& key, const Operands& operands) {
+  using Detail::convert;
+  auto bits = BitBuilder<riscv::unsigned32_t>(key["funct7"])
```
ErikP0: Is this intended, as the key["func7"] is again used later ?

ErikP0 (approved):
Looks good, although the tests of the assemble functions ("test/arch/riscv/assemble-test.cpp") do not assert if the assembled value is correct. This must be done prior to merging!
EDIT: And please make sure that all tests pass

include/arch/common/abstract-register-node.hpp
```diff
@@ -46,18 +51,14 @@ class RegisterNode : public AbstractSyntaxTreeNode {
    */
   ValidationResult validate(MemoryAccess& memoryAccess) const override;
 
-  const std::string& getIdentifier() const override;
-
   /**
-   * \return Assembled register identifier
+   * @return The identifier of the register.
```
janschopohl: \ 😉

include/arch/riscv/format.hpp
```diff
@@ -0,0 +1,206 @@
...
+ * The 6 formats implemented here are (as specified by the RISC-V standard):
+ *
+ * - R (register-register instructions, like `add`)
+ * - I (register-integer instructions, like `xori`)
+ * - S (store instructions)
+ * - SB (branch instructions)
+ * - U (LUI, AUIPC)
+ * - UJ (JAL)
+ */
+namespace Format {
```
janschopohl, line 53: 👍 the documentation of this file is really good

include/arch/riscv/load-store-instructions.hpp
```diff
@@ -155,7 +154,7 @@ class LoadInstructionNode
   }
 
   MemoryValue getValue(MemoryAccess& memoryAccess) const override {
-    assert(super::validate(memoryAccess));
+    assert::that(super::validate(memoryAccess));
```
janschopohl, line 158: I guess this file should include `common/assert.hpp` now?

include/arch/riscv/register-node-factory.hpp
```diff
@@ -18,34 +18,36 @@
 #ifndef ERAGPSIM_ARCH_RISCV_REGISTER_NODE_FACTORY_HPP
 #define ERAGPSIM_ARCH_RISCV_REGISTER_NODE_FACTORY_HPP
 
-#include <unordered_map>
+#include <string>
+#include <unordered_set>
 
 #include "arch/common/abstract-register-node-factory.hpp"
 
+class Archiitecture;
```
janschopohl: :wink:

include/arch/riscv/utility.hpp
```diff
@@ -102,9 +113,22 @@ void storeRegister(MemoryAccess& memoryAccess,
 template <typename UnsignedWord>
 std::enable_if_t<std::is_unsigned<UnsignedWord>::value, bool>
 isAddressValid(MemoryAccess& memoryAccess, UnsignedWord absoluteAdress) {
-    UnsignedWord lowerBound = 0;//TODO first adress of assembled code
-    UnsignedWord upperBound = memoryAccess.getMemorySize().get();//TODO last adress of assembled code
-    return absoluteAdress >= lowerBound && absoluteAdress < upperBound;
+  UnsignedWord lowerBound = 0;// TODO first adress of assembled code
+  UnsignedWord upperBound =
+      memoryAccess.getMemorySize().get();// TODO last adress of assembled code
```
janschopohl: why are there `TODO`s there?

source/arch/riscv/format.cpp
```diff
@@ -0,0 +1,155 @@
...
+      {"R", Format::R},
+      {"I", Format::I},
+      {"S", Format::S},
+      {"U", Format::U},
+      {"SB", Format::SB},
+      {"UJ", Format::UJ}
+  };
+  // clang-format on
+
+  assert::that(assemblers.count(format) > 0);
```
janschopohl, line 51: `#include "common/assert.hpp` `?

source/arch/riscv/register-node.cpp
```diff
@@ -13,19 +13,23 @@
...
-#include "arch/riscv/instruction-node.hpp"
-#include "core/memory-value.hpp"
+RegisterNode::RegisterNode(const std::string& identifier)
+: AbstractRegisterNode(identifier) {
+}
 
-using AssemblerFunction =
-    std::function<std::vector<bool>(InstructionKey, std::vector<MemoryValue>)>;
+MemoryValue RegisterNode::assemble() const {
+  // all register identifiers of riscv start with x
```
janschopohl: what about the `pc`? Is this a even concern here?

include/common/utility.hpp
```diff
@@ -498,44 +499,17 @@ std::vector<bool> convertToBinary(T value, std::size_t minSize = 0) {
...
+constexpr T discreteCeiling(const T &value, const S &divider) {
   return (value + divider - 1) / divider;
 }
 
 // Only for completeness.
 template <typename T, typename S = T>
-constexpr T divideFloor(const T& value, const S& divider) {
+constexpr T discreteFloor(const T &value, const S &divider) {
   return value / divider;
 }
```
janschopohl, line 512: these changes might cause the problems for the parser test.

janschopohl (approved):

include/common/utility.hpp
davschneller: Indeed, this is probably the reason why the parser tests currently fail, I think. (if the functions are reverted to the stuff they once were, it should be fine again)

source/arch/riscv/register-node.cpp
goldsborough: We should probably store the `RegisterInformation` and not the identifier only, to have the numeric ID. Or just the numeric ID

goldsborough: Because aliases

PR #192, opened 2016-12-16
Feature/protected memory runtime check

### LINKED ISSUE #154: Access to protected memory with runtime error
Any writing memory access (store, etc) causes a runtime error if the resulting address points into the protected memory area

k-lukas:
closes #154


janschopohl (approved):

ErikP0 (approved):

include/core/project.hpp
```diff
@@ -93,6 +93,21 @@ class Project : public Servant {
   MemoryValue setMemoryValueAt(size_t address, const MemoryValue &value);
 
   /**
+   * Calls Memory::isProtected(size_t, size_t)
+   */
+  bool isMemoryProtectedAt(std::size_t address, std::size_t amount = 1) const;
```
goldsborough: Why is this `std::size_t` here? I think you'd want to use just `size_t`

goldsborough (approved):

PR #199, opened 2016-12-19
Feature/gui input

### LINKED ISSUE #123: input-text

riedersa:
Closes #123 
I fixed the Problem with length_t in the qml components.


include/ui/input-text-model.hpp
```diff
@@ -25,7 +25,9 @@
 
 #include "core/memory-access.hpp"
 
-using length_t = unsigned ;
+namespace InputText{
+    using length_t = unsigned ;
```
goldsborough: clang format. Any reason this is inside a namespace and not just the class itself?

include/ui/input-text-model.hpp
```diff
@@ -25,7 +25,9 @@
...
+}
 
 class InputTextModel : QObject {
   Q_OBJECT
```
goldsborough, line 33: If it can't be in the class for some Qt reason, I would have a `using length_t = InputText::length_t;` here so you don't have to type `InputText' everywhere

goldsborough (approved):

include/ui/input-text-model.hpp
riedersa: I could not use InputTextModel::length_t in ui.cpp for updating qml. This way it works

PR #206, opened 2016-12-21
Feature/gui errors

### LINKED ISSUE #161: Better errors
Inline error messages

FelixOpolka:
closes #161
Improved gui for errors. Supports multiple inline errors per line and adaptive layout. Design may not be perfect yet (especially sidebar), but functionality is fine.


riedersa (approved):
Looks fine. I have not found any problem.

source/ui/Components/editor/Editor.qml
```diff
@@ -279,7 +279,9 @@ ScrollView {
                                                                                  "text": macro["code"]});
                         macroDisplayObject["subeditor"] = subeditor;
                         // Add triangle-button to display object
-                        var triangleButton = triangleButtonComponent.createObject(sidebar._errorBar, {"y": textArea.positionToRectangle(linePosition).y-2, "x": 0});
+                        var triangleButton = triangleButtonComponent.createObject(sidebar._macroBar, {"y": textArea.positionToRectangle(linePosition).y-3});
```
janschopohl: The hardcoded `-3` value might create Problems for different font sizes, I thinkt somethink relative to the font size/line height, for example `cursorRectangle.height/5` is more robust

include/ui/editor-component.hpp
```diff
@@ -153,7 +153,7 @@ class EditorComponent : public QObject {
   void deleteErrors();
 
   /** A signal to add an error in the editor. */
-  void addError(QString message, int line, QColor color);
+  void addIssue(QString message, int line, QString color);
```
janschopohl: make the QStrings `const QString&`

source/ui/Components/editor/Editor.qml
```diff
@@ -635,19 +732,19 @@ ScrollView {
...
                         }
 
                         Rectangle {
                             id: breakpointIcon
-                            height: Math.min(parent.height, errorBar.width) * 0.8
+                            height: textArea.cursorRectangle.height -1
                             width: height
                             anchors.verticalCenter: parent.verticalCenter
-                            anchors.left: parent.left
-                            anchors.leftMargin: height*0.15
```
janschopohl, line 662: Why did you change the height and anchors of this? It seems to work worse for me now, the breakpoints are a bit too big and sometimes clip with the left edge. The height with `Math.min` was to prevent them from being bigger then the height or width of the breakpoint bar.

source/ui/Components/editor/Editor.qml
```diff
@@ -563,70 +565,165 @@ ScrollView {
...
+
+                                horizontalAlignment: Text.AlignRight
+                                anchors.right: parent.right
                             }
-                            height: textArea.cursorRectangle.height
                         }
                     }
                 }
 
+                // Display errors, warnings and notes. Overlays lineNumberBar.
```
janschopohl: Weren't the breakpoints and errors supposed to be in a bar next to the linenumbers in order to prevent line numbers from being obstructed? I actually fixed this once in a review request.

source/ui/Components/editor/IssueMark.qml
```diff
@@ -0,0 +1,234 @@
...
+                // Displays the issueMessage (error message, warning message, information message).
+                Text {
+                    id: issueText
+                    anchors.right: parent.right
+                    anchors.rightMargin: _textMargin
+                    anchors.verticalCenter: parent.verticalCenter
+                    wrapMode: Text.WrapAtWordBoundaryOrAnywhere
+                    width: scrollView.width - sidebar.width - issueItemIcon.width - 3*_textMargin
+
+                    font.pixelSize: 10
```
janschopohl: I think hardcoding the font without reference to the editor font size will cause problems for some users. Also, the font seems really small on my system.

source/ui/Components/editor/IssueMark.qml
```diff
@@ -0,0 +1,234 @@
...
+                }
+
+                // Displays the issueMessage (error message, warning message, information message).
+                Text {
+                    id: issueText
+                    anchors.right: parent.right
+                    anchors.rightMargin: _textMargin
+                    anchors.verticalCenter: parent.verticalCenter
+                    wrapMode: Text.WrapAtWordBoundaryOrAnywhere
+                    width: scrollView.width - sidebar.width - issueItemIcon.width - 3*_textMargin
```
janschopohl: I don't know exactly what the problem is, but the the error text width doesn't properly update when zooming the editor. Look at the editor components for scaling, I think replacing `scrollView.width` by [this](https://github.com/TUM-LRR/era-gp-sim/blob/ec6ba63d8ca9f5eece8444551bd8d8ed817daffb/source/ui/Components/editor/Editor.qml#L154) might help. But I have a feeling something else has to be changed, too 😉

janschopohl (approved):

PR #214, opened 2016-12-25
Fixed output console not being visible for some users.

janschopohl:
Please check if the output console works correctly now.

### DIFF
--- a/source/ui/Components/Output/Console.qml
+++ b/source/ui/Components/Output/Console.qml
@@ -7,17 +7,15 @@ Item {
     property int outputItemIndex: 2
     property int oldMode: 0
     property var oldAddress: 0
-    id: it
+    id: item
 
     TextArea {
         id: textarea
-        height: parent.height
-        width: parent.width
-        anchors.top: parent.top
-        anchors.left: parent.left
+        anchors.fill: parent
         style: TextAreaStyle {
             textColor: "white"
             backgroundColor: "black"
+            renderType: Text.QtRendering
         }
         wrapMode: Text.Wrap
         readOnly: true
@@ -48,11 +46,11 @@ Item {
             var _mode = outputComponent.getOutputItem(outputItemIndex)["textMode"];
             // Check if the memory address that was changed (at least partly) belongs to
             // the output item's source space.
-            if ((address+length) >= _baseAddress && _mode === 0) {
-                it.updateContent(_baseAddress);
-            }else if(_mode !== 0 ){
-                if(address<=_baseAddress && address+length >= _baseAddress){
-                    it.updateContent(_baseAddress);
+            if ((address + length) >= _baseAddress && _mode === 0) {
+                item.updateContent(_baseAddress);
+            } else if (_mode !== 0 ){
+                if(address <= _baseAddress && address + length >= _baseAddress) {
+                    item.updateContent(_baseAddress);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


source/ui/Components/Output/Console.qml
```diff
@@ -1,75 +1,73 @@
...
-            id: textarea
-            height: parent.height
-            width: parent.width
-            anchors.top: parent.top
-            anchors.left: parent.left
-
-            wrapMode: Text.Wrap
+    TextArea {
+        id: textarea
+        height: parent.height
```
andizimmerer: Consider using `anchors` over property bindings as anchors are implemented more efficient in qml. Or just use `anchors.fill: parent`

source/ui/Components/Output/Console.qml
```diff
@@ -1,75 +1,73 @@
...
+        // Send when the memory changes (at any address).
+        onMemoryChanged: {
+            var _baseAddress = outputComponent.getOutputItem(outputItemIndex)["baseAddress"];
+            var _mode = outputComponent.getOutputItem(outputItemIndex)["textMode"];
+            // Check if the memory address that was changed (at least partly) belongs to
+            // the output item's source space.
+            if ((address+length) >= _baseAddress && _mode === 0) {
+                it.updateContent(_baseAddress);
+            }else if(_mode !== 0 ){
+                if(address<=_baseAddress && address+length >= _baseAddress){
```
andizimmerer: maybe add some spaces between operators ;)

source/ui/Components/Output/Console.qml
```diff
@@ -1,75 +1,73 @@
...
-
-            wrapMode: Text.Wrap
+    TextArea {
+        id: textarea
+        height: parent.height
+        width: parent.width
+        anchors.top: parent.top
+        anchors.left: parent.left
+        style: TextAreaStyle {
             textColor: "white"
```
andizimmerer, line 16: Default font size is very small. Maybe someone could add zoom

source/ui/Components/Output/Console.qml
```diff
@@ -1,75 +1,73 @@
...
+            acceptedButtons: Qt.RightButton
+            onDoubleClicked: {
+                if (outputComponent.getOutputItem(outputItemIndex)["textMode"] !== 0) {
+                    textarea.clear();
+                }
             }
-
         }
+        function clear() {
+            text = "";
```
andizimmerer, line 37: Interesting: writing a `\0` to memory in pipe mode also clears the console. Is this intended behaviour or is it a bug?

source/ui/Components/Output/Console.qml
```diff
@@ -1,75 +1,73 @@
...
+        wrapMode: Text.Wrap
+        readOnly: true
+        backgroundVisible: true
+
+        //Clears the Screen in pipelike mode
+        MouseArea {
+            id: clearArea
+            anchors.fill:parent
+            visible: true
+            acceptedButtons: Qt.RightButton
```
andizimmerer, line 29: looks like a hidden feature 😅 
Maybe you could add a small hint in the bottom right corner?
It might be more obvious to use Ctrl+C?

source/ui/Components/Output/Console.qml
```diff
@@ -1,75 +1,73 @@
...
+            var baseAddress = outputComponent.getOutputItem(outputItemIndex)["baseAddress"];
+            if(mode === oldMode && baseAddress === oldAddress){
+                return;
             }
+            it.updateContent(outputComponent.getOutputItem(outputItemIndex)["baseAddress"]);
+            settingsWindowC.updateSettings();
         }
     }
+
     function updateContent(_baseAddress) {
```
andizimmerer, line 69: Try out some characters (decimal code): 
- 27: Escape "" (greater equal sign)
- 7: Bell "" (°)
- 8: Backspace "" (plus-minus sign)
I am not sure if this has something to do with your code. In fact the correct ASCII values are returned but they are displayed the wrong way on my machine. Maybe this has something to do with the default font?

janschopohl: hmm this was actually just a small fix (~10 lines) about setting the background of the console, everything else was just formatting. So I guess this bug should be in a separate issue (if it really is a bug).

source/ui/Components/Output/Console.qml
janschopohl: see comment below

janschopohl:
I created #233 for the comments @Jibbow made about the console features/bugs, as this issue is just about adding the right background on all maschines.

goldsborough (approved):

source/ui/Components/Output/Console.qml
andizimmerer: oh sorry, all these changes were shown to me on this branch. There is probably some other branch included which is not already on master ;)

PR #219, opened 2016-12-26
Integrated condition variable into scheduler and changed command queue implementation

### LINKED ISSUE #218: Improve scheduler performance

janschopohl:
closes #218


janschopohl:
Temporarily closed as I had a new idea how to improve this by syncing the ui.

source/core/scheduler.cpp
```diff
@@ -42,15 +50,19 @@ void Scheduler::_run() {
...
-    Task task;
-    if (_taskQueue.pop(task)) {
-      // there was something in the queue, ececute the task
-      task();
-
-    } else {
-      // the queue is empty, sleep for a bit
-      std::this_thread::sleep_for(_sleepMilliseconds);
+    std::unique_lock<std::mutex> lock(_mutex);
+    while (_taskQueue.empty()) {
```
goldsborough: Couldn't we use the predicate feature of the condition variable here?
```
_conditionVariable.wait(lock, [&_taskQueue]{ return !_taskQueue.empty(); })
```

source/core/scheduler.cpp
```diff
@@ -42,15 +50,19 @@ void Scheduler::_run() {
...
+      // wait while there is nothing in the queue, guards for spurious wakeups.
+      _conditionVariable.wait(lock);
     }
+    // make sure the mutex is locked and pop an element from the queue (which is
+    // guaranteed to not be empty)
+    if (!lock) lock.lock();
```
goldsborough: You make these checks here (and in other places), but one of the properties of the condition variable is that the mutex is reacquired upon wakeup. As such I think this check is unnecessary?

tests/core/project-test.cpp
```diff
@@ -89,6 +89,9 @@ class ProjectTestFixture : public ::testing::Test {
     MemoryManager memoryManager = projectModule.getMemoryManager();
     memoryManager.setErrorCallback(errorCallback);
     architectureValidator.validate();
+    CommandInterface commandInterface = projectModule.getCommandInterface();
+    commandInterface.setSyncCallback(
+        [this]() { this->projectModule.guiReady(); });
```
goldsborough: There's special syntax that let's you write a lambda without arguments simply as `[...]{...}` (no parantheses)

goldsborough (changes requested):

PR #222, opened 2016-12-26
Hotfix/parser refactor

### LINKED ISSUE #168: Refine SymbolTable, General Parser Refactoring
https://github.com/TUM-LRR/era-gp-sim/tree/hotfix/parser-refactor

davschneller:
closes #168

Note: The documentation might still be incomplete...
Ok, this branch basically does change a bit the parser internally and externally. In short:

* `CompileState` does not exist any more (it was a class with many different not-connected purposes, in my opinion not ideal), so it has been replaced by a `CompileErrorList` and several others.
* The position of arguments etc. is not propagated through the whole parser allowing character-accurate error message placing (thought currently only line-wise placement is supported by the gui).
* The parser has now been split up into smaller sub modules:
  * `era-sim-parser-common` contains some generally used methods, not depending on any other parser module, so it carries low dependencies to the outer world (e.g. `era-sim-arch-common`).
  * `era-sim-parser-factory` contains the factory which creates the different parsers. It has been singled out from the common module, because it depends on all the specific implemented parsers. (does this make sense? And also, is there a possibility in cmake to link to, for example, all targets of a sub folder?)
  * `era-sim-parser-riscv` contains everything that is only RISC-V specific.
  * `era-sim-parser-independent` contains classes which are related to the RISC-V parser, but might not be restricted to it. That is the reason, they all got into a separate module.
* Turned many `structs` (but not all) into `classes`.
* This also encorporated some small changes in the external interface.
* Several small bug fixes.
* Various smaller changes to hopefully resemble a more C++-like coding style.

It might be nice, if a few of you might look over this branch, as it is, in my optinion, quite large. Also, some system tests would probably be nice (n > 2 eyes see more than two alone of course)
Please tell me, if there is something that I should fix.

Bye!


include/parser/independent/enhance-symbol-table-immutable-arguments.hpp
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * \brief The memory allocator used to convert relative to absolute positions
+   * and also for different section management.
+   */
+  MemoryAllocator _allocator;
+};
+
+#endif /* ERAGPSIM_PARSER_INDEPENDENT_ENHANCE_SYMBOL_TABLE_IMMUTABLE_ARGUMENTS_HPP \
+        * \                                                                        \
+        * \ \                                                                                 \
```
kreudom: I assume this comment was an accident

include/parser/independent/intermediate-operation.hpp
```diff
@@ -0,0 +1,228 @@
...
+#include "parser/common/code-position-interval.hpp"
+#include "parser/independent/positioned-string.hpp"
+
+class Architecture;
+class SymbolReplacer;
+class MemoryAllocator;
+
+class MemoryValue;
+#include "parser/common/final-command.hpp"
+class FinalCommand;
```
kreudom: This forward declaration isn't really necessary with the include right above.

kreudom (commented):
Hi, good work on this!

Additionally to the two things I noted, I noticed the define guard names are inconsistent in most files now. Some include the full file path, some only parser + file name and some only the class name.

But these are all very small problems. All in all this feature looks very good imo.

kreudom (approved):

kreudom (approved):

davschneller:
Thanks for the review! All requested changes should be implemented by now.

ErikP0:
Hi, 
first of all: What a big feature! You did great work, the parser feels tidied up now, I like it!
I have a number of **cosmetic** suggestions.
1. In compile-error.hpp l.87: remove the const qualifier of return type of severity().
Because enums are essentially integer values, the const qualifier for return type on a non-class types is redundant as they cannot be changed (even without const)
[see SO](http://stackoverflow.com/questions/1607188/why-is-a-type-qualifier-on-a-return-type-meaningless)

2. In macro-information.hpp l.65: old, unused code?
3. In expression-compiler-definitions.hpp l.61: missing documentation on the CONSTANT enum (What's the difference between constant and literal?)
4. In intermediate-instruction.hpp l.139, l.151: consider a using declaration for pleasant reading
5. In intermediate-macro-instruction l.132: TODO? `TODO?`
6. In intermediate-operation.hpp l.223: old, unused code?
7. In macro-directive.hpp l.128: old,unused code?
8. In memory-definition-directive.hpp l.97: As this is the message for memory definition (not reservation), I recommend to use a slightly different message: e.g. `"define static memory"`
9. In positioned-string: Have you considered moving this class into parser/common?
10. In symbol-graph-evaluation.hpp, symbol-graph.hpp, symbol-replacer: please add some documentation on the purpose of these classes. The documentation of the used algorithms is very good, though.
11. In symbol.hpp l.45: something more important must have happened ;)
12. In syntax-tree-generator.hpp l. 45: consider a using declaration for pleasant reading

ErikP0 (approved):

include/common/multiregex.hpp
```diff
@@ -44,6 +44,24 @@ struct MultiregexMatch {
...
 
   /**
+   * \brief Creates a new multiregex match with the given arguments.
+   * \param source The COMPLETE source string (the relevant part will be cut
+   * out).
+   * \param position The positition of the match in the string.
+   * \param length The length of the match.
+   * \param choice The taken choice.
+   */
+  MultiregexMatch(const String& source = "",
```
goldsborough: Default it to `String()`

include/parser/common/compile-error-list.hpp
```diff
@@ -0,0 +1,206 @@
...
+                           (interval),                                       \
+                           QT_TRANSLATE_NOOP("Parser Information", message), \
+                           ##__VA_ARGS__)
+/**
+ * \brief A list of compile errors, enhanced by some helper methods.
+ */
+class CompileErrorList {
+ public:
+  /**
+   * \brief Constructs an empty compile error list.
```
goldsborough: Please remove all `\brief`s. The first line is implicitly `\brief`

include/parser/common/compile-error-list.hpp
```diff
@@ -0,0 +1,206 @@
...
+   */
+  CompileErrorList() = default;
+
+  /**
+   * \brief The internal compile error vector.
```
goldsborough: For getters it's ok to just use `\return` without the `\brief` section

include/parser/common/compile-error-list.hpp
```diff
@@ -0,0 +1,206 @@
...
+   * \return The internal compile error vector.
+   */
+  const CompileErrorVector errors() const noexcept;
```
goldsborough: I think this should be a `const&`

include/parser/common/compile-error-list.hpp
```diff
@@ -0,0 +1,206 @@
...
+   * \return True, if the vector is empty, else false.
+   */
+  bool empty() const;
+
+  /**
+   * \brief A helper method to determine if the number of errors (not including
+   * warnings, information).
+   * \return The number of errors.
+   */
+  std::size_t errorCount() const;
```
goldsborough: `using size_t = std::size_t`

include/parser/common/compile-error-list.hpp
```diff
@@ -0,0 +1,206 @@
...
+   * \return The internal compile error vector.
+   */
+  const CompileErrorVector errors() const noexcept;
+
+  /**
+   * \brief A helper method to determine if the list has any errors (not
+   * including warnings, information).
+   * \return True, if errors exist, else false.
+   */
+  bool hasErrors() const;
```
goldsborough: All these getters can be `noexcept`

include/parser/common/compile-error-list.hpp
```diff
@@ -0,0 +1,206 @@
...
+   * \param message The message to record. (do not change to std::string as
+   * input parameter! We want to force that people insert a string literal here,
+   * for translation using the QT_TRANSLATE_NOOP macro).
+   * \param parameters The parameters for the message.
+   */
+  template <typename... Args>
+  void pushCompileErrorInternal(CompileErrorSeverity severity,
+                                const CodePositionInterval& interval,
+                                const char* message,
+                                const Args&... parameters) {
```
goldsborough: It doesn't matter here, but in general, when forwarding arguments in c++, it's a good idea to use perfect forwarding. See [here](http://eli.thegreenplace.net/2014/perfect-forwarding-and-universal-references-in-c/)

include/parser/common/compile-error-list.hpp
```diff
@@ -0,0 +1,206 @@
...
+    auto error =
+        CompileError(std::make_shared<Translateable>(message, parameters...),
```
goldsborough: You could just write `addRaw({ ... })` and put the arguments in the `...`

include/parser/common/compile-error-list.hpp
```diff
@@ -0,0 +1,206 @@
...
+                     interval,
+                     severity);
+    addRaw(error);
+  }
+
+  /**
+   * \brief Adds a compile error created from the arguments to the list,
+   * pre-specified of severity 'error'.
+   * \param interval The given code position interval where to error occured.
```
goldsborough, line 157: One newline between `\brief` and the rest.

include/parser/common/compile-error-list.hpp
```diff
@@ -0,0 +1,206 @@
...
+  void pushInformationInternal(const CodePositionInterval& interval,
+                               const char* message,
+                               const Args&... parameters) {
+    pushCompileError(
+        CompileErrorSeverity::INFORMATION, interval, message, parameters...);
+  }
+
+ private:
+  /**
+   * \brief The internal vector for compile errors.
```
goldsborough: It's ok to write this on one line: /** The internal vector for compile errors. */

include/parser/common/code-position-interval.hpp
```diff
@@ -0,0 +1,146 @@
...
+   */
+  CodeCoordinate endCharacter() const noexcept;
+
+  /**
+   * \brief Determines, if the code position interval is empty.
+   * \return True, if either the line index of the last position is smaller than
+   * the index of the first line or the start and last line are the same and the
+   * x-indices are reversed.
+   */
+  bool empty() const noexcept;
```
goldsborough: Prefer `is_empty` here

include/parser/common/code-position-interval.hpp
```diff
@@ -0,0 +1,146 @@
...
+   */
+  CodePositionInterval unite(const CodePositionInterval& other) const;
+
+  /**
+   * \brief Builds the biggest interval containing an area contained within two
+   * code position intervals.
+   * \param other The code position interval to cut with.
+   * \return The resulting interval.
+   */
+  CodePositionInterval cut(const CodePositionInterval& other) const;
```
goldsborough: I think you mean `intersect` haha

include/parser/independent/constant-directive.hpp
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * \brief Executes the constant directive, i.e. checks its validity.
+   * \param immutable Some constant arguments which might be helpful.
+   * \param errors The compile error list to note down any errors.
+   * \param commandOutput The final command output vector to record all
+   * finalized commands.
+   * \param memoryAccess The memory access used to reserve memory and validate
+   * instructions.
+   */
+  virtual void execute(const ExecuteImmutableArguments& immutable,
```
goldsborough, line 56: You don't need both `virtual` and `override`

include/parser/independent/symbol-replacer.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include <functional>
+#include <vector>
+#include "common/multiregex.hpp"
+#include "parser/independent/positioned-string.hpp"
+#include "parser/independent/symbol.hpp"
+class SymbolGraphEvaluation;
+class CompileErrorList;
+
+class SymbolReplacer {
```
goldsborough, line 34: This class could use some more documentation :)

include/parser/independent/symbol-graph.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#ifndef ERAGPSIM_PARSER_INDEPENDENT_SYMBOL_GRAPH_HPP
+#define ERAGPSIM_PARSER_INDEPENDENT_SYMBOL_GRAPH_HPP
+
+#include <string>
+#include <vector>
+#include "parser/independent/symbol.hpp"
+class SymbolGraphEvaluation;
+
+class SymbolGraph {
```
goldsborough, line 32: Docs

include/parser/independent/symbol-graph.hpp
```diff
@@ -0,0 +1,57 @@
...
+ public:
+  using size_t = std::size_t;
+  SymbolGraph() = default;
```
goldsborough: You don't need to declare the default constructor if you haven't defined any other constructors

include/parser/independent/symbol-graph.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+  void addNode(const Symbol& symbol);
+
+  SymbolGraphEvaluation evaluate() const;
+
+  class SymbolNode {
```
goldsborough, line 39: Please put nested types at the top of the class (and document stuff)

ErikP0:
I'm sorry. There is one thing that needs attention before this feature can be merged:
I cannot get backwards jumps working, like
```
loop:
add x0, x0, x0
jal x0, loop
```
Gives me a compile error for `jal` stating `Integer out of range`.
Some debugging deduced the following: `ImmediateFactory::labelToImmediate(...)` returns a 64bit MemoryValue that is the relative address for each instruction & label (like loop). This MemoryValue is converted into a hex string (e.g. "0xfffffffffffffffe"). This hex string is later parsed by the `integer-parser` class with `std::stoi`. Number too big (as stoi is for int) -> exception -> compile error.
So far, so good.
Changing `ImmediateFactory::labelToImmediate(...)` to return a 32bit MemoryValue unfortunately does not resolve this problem. The hex string is correct (e.g. "0xfffffffe") but for whatever reason `std::stoi` still throws a `out_of_range` exception that leads to a compile error.
Try this:
```
std::string s = "0xfffffffe";
int i = std::stoi(s, nullptr, 16);
```
We may need to find a solution in the hangout on monday, if you have no idea how to parse the hex string the right way.

source/parser/independent/symbol-graph.cpp
```diff
@@ -0,0 +1,303 @@
...
+#include "parser/independent/positioned-string.hpp"
+#include "parser/independent/symbol-graph-evaluation.hpp"
+
+using size_t = std::size_t;
+
+void SymbolGraph::addNode(const Symbol& symbol) {
+  // Adding a node means checking all adjencencies to it and from it away – and
+  // also to itself.
+  auto newNode = SymbolNode(symbol, _nodes.size());
+  for (auto& node : _nodes) {
```
goldsborough: `const auto&`

source/parser/independent/symbol-graph.cpp
```diff
@@ -0,0 +1,303 @@
...
+    newNode.checkAdjacency(node);
+  }
+
+  newNode.checkAdjacency(newNode);
+
+  _nodes.push_back(newNode);
+}
+
+// Some small helper classes.
+class PartialEvaluation {
```
goldsborough, line 31: Put helper classes/functions like this into an anonymous namespace

source/parser/independent/symbol-graph.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+struct StackEntry {
+  size_t index;
+  size_t position;
+
+  StackEntry(size_t index, size_t position) : index(index), position(position) {
+  }
+};
+
+static std::vector<size_t>
```
goldsborough: You can get rid of the `static` once you've put all this into an anonymous namespace

source/parser/independent/symbol-graph.cpp
```diff
@@ -0,0 +1,303 @@
...
+decomposeCycle(size_t startNode, std::stack<StackEntry>& stack) {
+  // Here, we decompose the stack to find a cycle.
+  std::vector<size_t> cycle;
+  cycle.push_back(startNode);
+
+  // We know, (at least: when called in this one place in the function below)
+  // that 'startNode' exists again in the stack, so we go down, until we find
+  // it.
+  assert::that(!stack.empty());
```
goldsborough: Put this at the top of the function

source/parser/independent/symbol-graph.cpp
```diff
@@ -0,0 +1,303 @@
...
+  // and did some crazy constant stuff, the whole system would not crash because
+  // of out of stack... But this might also be thought a bit too far... But hey,
+  // now we've already got it coded!)
+
+  // The stack for iterative DFS (works like described on Wikipedia).
+  std::stack<size_t> toVisit;
+
+  // Some means to emulate the call stack.
+  std::stack<StackEntry> emulatedStack;
+  std::vector<bool> fastLookup(nodes.size());
```
goldsborough: call this `currentPath` or something

source/parser/independent/symbol-graph.cpp
```diff
@@ -0,0 +1,303 @@
...
+static std::vector<size_t>
+connectedDfs(const std::vector<SymbolGraph::SymbolNode>& nodes,
+             std::vector<size_t>& topologicOrder,
+             std::vector<bool>& visited,
+             size_t start) {
+  // Now here comes the fun part (almost as fun as implementing the Hierholzer
+  // algorithm for Euler cycles iteratively, it's horrible), we do iterative DFS
+  // but with checking for a DAG and generating of finishing numbers for each
+  // node (we need it for the topologic sort). Sounds fun? So it is. Much fun.
+  // (why iterative? Well, the main reason was that if some day some maniac came
```
goldsborough: What `constant` stuff do you mean? You'd need to use a lot of memory to get a stack overflow. Most graph algorithms are implemented recursively. I would say "it's ok, because it's a fun challenge to do it iteratively". However it also makes it more difficult to understand, since graph algorithms like this one are naturally recursive. It might be difficult for future maintainers, since anyway many people are not *that* comfortable with graph algorithms

source/parser/independent/symbol-graph.cpp
```diff
@@ -0,0 +1,303 @@
...
+      return decomposeCycle(node, emulatedStack);
+    }
+
+    // In the end, we need to emulate the recursive backtracking on our virtual
+    // stack. For this, we save the number of items on our toVisit stack when we
+    // put our node on the stack. If we are back down, it means that we are back
+    // at the old junction again and our node is not visited any more.
+    while (!emulatedStack.empty() &&
+           emulatedStack.top().position == toVisit.size()) {
+      // If so, we have finished a node, so we put it to the (inverse) topologic
```
goldsborough: `s/inverse/reverse`

source/parser/independent/symbol-graph.cpp
```diff
@@ -0,0 +1,303 @@
...
+    // And all of that, we do over and over again.
+  }
+
+  // If we have finished without problems, there has not been a cycle.
+  return std::vector<size_t>();
+}
+
+static PartialEvaluation
+disconnectedDfs(const std::vector<SymbolGraph::SymbolNode>& nodes) {
+  // To check for cycles, we to a DFS over the graph which might not be
```
goldsborough: `s/to/do`

source/parser/independent/symbol-graph.cpp
```diff
@@ -0,0 +1,303 @@
...
+  // connected...
+  std::vector<bool> visited(nodes.size());
+  std::vector<size_t> topologicOrder;
+
+  // So we have to go over all the nodes until everyone is visited.
+  for (const auto& node : Utility::range<size_t>(0, nodes.size())) {
+    if (!visited[node]) {
```
goldsborough: I prefer `if (visited[node]) continue;` to then have one level less of indentation

source/parser/independent/symbol-graph.cpp
```diff
@@ -0,0 +1,303 @@
...
+    // put our node on the stack. If we are back down, it means that we are back
+    // at the old junction again and our node is not visited any more.
+    while (!emulatedStack.empty() &&
+           emulatedStack.top().position == toVisit.size()) {
+      // If so, we have finished a node, so we put it to the (inverse) topologic
+      // order.
+      auto index = emulatedStack.top().index;
+      emulatedStack.pop();
+      fastLookup[index] = false;
+      topologicOrder.push_back(index);
```
goldsborough: You can probably replace all calls to `push_back` with `emplace_back` (everwhere)

source/parser/independent/symbol-graph.cpp
```diff
@@ -0,0 +1,303 @@
...
+  return _symbol;
+}
+size_t SymbolGraph::SymbolNode::index() const noexcept {
+  return _index;
+}
+const std::vector<size_t>& SymbolGraph::SymbolNode::adjacent() const noexcept {
+  return _adjacent;
+}
+
+void SymbolGraph::SymbolNode::checkAdjacency(const SymbolNode& other) {
```
goldsborough: I think the name `checkAdjacency` is a bit misleading. I was expecting a function that tells me if vertex `v` is connected to vertex `u`. This does something else. Maybe rename it to`connectIf<whatever_you_are_trying_to_check_here>`

source/parser/independent/symbol-replacer.cpp
```diff
@@ -0,0 +1,121 @@
...
+      return _replacer(symbol);
+    } else {
+      return symbol.value().string();
+    }
+  };
+
+  // Then iterate at max 'maximumReplaceCount' round.
+  for (const auto& round :
+       Utility::range<std::size_t>(0, _maximumReplaceCount)) {
+    (void)round;// (to prevent unused variable warnings)
```
goldsborough: In Python we name unused variables `_`. It's not going to satisfy the C++ compiler, but may make your intent clearer

source/parser/independent/symbol.cpp
```diff
@@ -0,0 +1,65 @@
...
+* You should have received a copy of the GNU General Public License
+* along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "parser/independent/symbol.hpp"
+#include "common/assert.hpp"
+
+// Some testing regexes.
+static const std::regex TRIMMED =
+    std::regex("(^\\s+|\\s+$)", std::regex_constants::optimize);
```
goldsborough: Maybe use `R"(^\\s+|\\s+$)"`? This will not surround the regex with parentheses, but you don't need those here anyway

source/parser/riscv/riscv-regex.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+RiscvParser::RiscvRegex::RiscvRegex() {
+}
+
+static size_t trimRight(const std::string &line, size_t pos) {
+  --pos;
+  while (pos > 0 && std::isspace(line[pos])) {
+    --pos;
+  }
+  ++pos;
```
goldsborough: return `++pos`

source/parser/riscv/riscv-regex.cpp
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// Returns true if successful
+bool RiscvParser::RiscvRegex::_readInstructionOrLabel(
+    const std::string &line,
+    CodeCoordinate lineCoordinate,
+    CompileErrorList &errors,
+    size_t &pos) {
+  // Skip spaces
+  for (; pos < line.size() && std::isspace(line[pos]); pos++) {
```
goldsborough: I'd prefer turning this into a while loop with the `++pos` as the body

source/parser/riscv/riscv-regex.cpp
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  size_t startPos = pos;
+  // Add valid characters to _instruction
+  for (; pos < line.size() &&
```
goldsborough: Again, while loop

include/parser/riscv/riscv-regex.hpp
```diff
@@ -0,0 +1,125 @@
...
+   * \param errors CompileErrorList to record errors.
+   * \return The last read position in the string (if the line does not match,
+   * this is the character that caused the error)
+   */
+  size_t matchLine(const std::string &line,
+                   CodeCoordinate lineCoordinate,
+                   CompileErrorList &errors);
+
+ private:
+  bool _readInstructionOrLabel(const std::string &line,
```
goldsborough, line 101: Private methods should also have documentation, for future developers and yourself

include/parser/riscv/riscv-regex.hpp
```diff
@@ -0,0 +1,125 @@
...
+#include <regex>
+#include <string>
+
+#include "parser/riscv/riscv-parser.hpp"
+
+class CompileErrorList;
+#include "parser/independent/positioned-string.hpp"
+
+/**
+ * Wrapper class for the RegExp used for parsing.
```
goldsborough, line 30: This class looks like it does its own string processing/matching. Why not use regular expressions?

scripts/includeguards.py
```diff
@@ -0,0 +1,29 @@
+import common
```
goldsborough, line 4: `#!/usr/bin/env python`

source/parser/independent/constant-directive.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+void ConstantDirective::execute(const ExecuteImmutableArguments& immutable,
+                                CompileErrorList& errors,
+                                FinalCommandVector& commandOutput,
+                                MemoryAccess& memoryAccess) {
+  // Try to parse argument to catch errors early.
+  auto fullExpression = immutable.replacer().replace(_expression, errors);
+  if (!fullExpression.string().empty()) {
+    immutable.generator().transformOperand(
+        _expression, immutable.replacer(), errors);
```
goldsborough, line 47: You can just `return` here and don't need an `else` block

source/parser/independent/expression-tokenizer.cpp
```diff
@@ -0,0 +1,87 @@
...
+std::vector<std::string> ExpressionTokenizer::buildRegexVector(
+    const std::vector<ExpressionTokenDefinition>& definitions) {
+  std::vector<std::string> output;
+  output.reserve(definitions.size() + 1);
+
+  // We copy the regexes from our definitions and also include a whitespace
+  // finder.
+  output.push_back("\\s+");
+  _typeMapping.reserve(definitions.size());
+  for (const ExpressionTokenDefinition& i : definitions) {
```
goldsborough: auto?

source/parser/independent/intermediate-instruction.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+std::vector<std::shared_ptr<AbstractSyntaxTreeNode>>
+IntermediateInstruction::compileArgumentVector(
+    const std::vector<PositionedString>& vector,
+    const ExecuteImmutableArguments& immutable,
+    CompileErrorList& errors,
+    MemoryAccess& memoryAccess) {
+  std::vector<std::shared_ptr<AbstractSyntaxTreeNode>> output;
+
+  SymbolReplacer::DynamicReplacer labelReplacerFunction = [&](
```
goldsborough, line 77: It's better to explicitly capture the variables you need. For example, you're capturing the `vector` here, but not using it. So `[&immutable]` will do

source/parser/independent/intermediate-instruction.cpp
```diff
@@ -0,0 +1,259 @@
...
+  }
+  return output;
+}
+
+FinalCommand IntermediateInstruction::compileInstruction(
+    const ExecuteImmutableArguments& immutable,
+    CompileErrorList& errors,
+    MemoryAccess& memoryAccess) {
+  // We replace all occurenced in target in source (using a copy of them).
+  auto srcCompiled =
```
goldsborough: I'd prefer `sourceCompiled` and `targetCompiled` or (or `compiledSource/Target`)

source/parser/independent/intermediate-instruction.cpp
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  // For now. Later to be reworked with a bit-level memory allocation?
+  auto instructionLength =
+      instructionSet[nameString].getLength() / architecture.getByteSize();
+  _relativeAddress = allocator["text"].allocateRelative(instructionLength);
+}
+
+static bool isWordCharacter(char c) {
+  return (c == '_' || std::isalpha(c) || std::isdigit(c));
```
goldsborough: `std::isalnum`

source/parser/independent/intermediate-instruction.cpp
```diff
@@ -0,0 +1,259 @@
...
+  replaceInVector(_sources, name, value);
+  replaceInVector(_targets, name, value);
+}
+
+std::string IntermediateInstruction::toString() const {
+  auto str = name().string();
+
+  // Append targets
+  for (auto i : Utility::range<size_t>(0, _targets.size())) {
+    if (i != 0) {
```
goldsborough: Better to use the positive condition in the `if` and the negative in the else

source/parser/independent/intermediate-instruction.cpp
```diff
@@ -0,0 +1,259 @@
...
+}
+
+void IntermediateInstruction::insertIntoArguments(
+    const PositionedString& name, const PositionedString& value) {
+  replaceInVector(_sources, name, value);
+  replaceInVector(_targets, name, value);
+}
+
+std::string IntermediateInstruction::toString() const {
+  auto str = name().string();
```
goldsborough, line 209: Consider caching this operation if it's called often!

source/parser/independent/intermediate-instruction.cpp
```diff
@@ -0,0 +1,259 @@
...
+    }
+    str += _sources[i].string();
+  }
+
+  str += "\n";
+  return str;
+}
+
+IntermediateOperationPointer IntermediateInstruction::clone() {
+  return IntermediateOperationPointer{new IntermediateInstruction{*this}};
```
goldsborough: `make_shared`

source/parser/independent/intermediate-macro-instruction.cpp
```diff
@@ -0,0 +1,148 @@
...
+    IntermediateOperationVectorIterator begin,
+    IntermediateOperationVectorIterator end,
+    MacroDirectiveTable& macroTable,
+    CompileErrorList& errors) {
+  for (auto i = begin; i != end; ++i) {
+    // Try casting to IntermediateInstruction, skip instruction if it doesn't
+    // work
+    if ((*i)->getType() != IntermediateOperation::Type::INSTRUCTION) continue;
+
+    IntermediateInstruction& inst = static_cast<IntermediateInstruction&>(**i);
```
goldsborough: I would normally have said `dynamic_cast`, but since you have the check above it's fine. However, this is a perfect use case for `auto`, since it's clear the result of casting to `IntermediateInstruction&` is `IntermediateInstruction&`

goldsborough (approved):
Helping my browser

davschneller:
Thank you very much for all of your reviews!
The issue should now contain almost all suggested changes, and more documentation. When Travis passes, I will merge it.

PR #223, opened 2016-12-28
Feature/gui input pipelike

### LINKED ISSUE #211: add pipelike mode to input

riedersa:
closes #211


include/ui/input-text-model.hpp
```diff
@@ -66,17 +72,35 @@ class InputTextModel : QObject {
...
+   * \return the length
+   */
+  Q_INVOKABLE int getMode();
+
  private:
   QQmlContext* _context;
   int _start;
   length_t _maximumLength;
   MemoryAccess _memoryAccess;
+  int _mode;
```
janschopohl: There should be comments for every private member. You should add an `enum class` for mode selection (you have to register the enum if you want to use it in qml), using different Integer values is extremely error prone.

include/ui/input-text-model.hpp
```diff
@@ -66,17 +72,35 @@ class InputTextModel : QObject {
...
+
+  /**
+   * \brief Gets the mode
+   * \return the length
+   */
+  Q_INVOKABLE int getMode();
+
  private:
   QQmlContext* _context;
   int _start;
```
janschopohl: I don't know exaclty what this does, as there is no comment, but I suspect that using `std::size_t` instead of `int` would be better

include/ui/input-text-model.hpp
```diff
@@ -43,6 +43,12 @@ class InputTextModel : QObject {
   Q_INVOKABLE void newText(QString text);
 
   /**
+   * \brief Sets a number in the memory, used for left/up/etc.
+   * \param number the number
+   */
+  Q_INVOKABLE void newNumber(int number);
```
janschopohl: use `std::size_t` instead of `int` here. Consider adding `using size_t = std::size_t` in this class.

source/ui/input-text-model.cpp
```diff
@@ -25,7 +25,8 @@ InputTextModel::InputTextModel(QQmlContext* context, MemoryAccess memoryAccess)
 , _context(context)
 , _start(0)
 , _maximumLength(20)
-, _memoryAccess(memoryAccess) {
+, _memoryAccess(memoryAccess)
+, _mode(0) {
   _context->setContextProperty("inputtextMod", this);
```
janschopohl: This didn't change here, but `inputtextMod` is not really a great name. `inputTextModel` would be better (styleguide). (sorry for the nitpicking, but I was confused for a short moment when I saw this name in qml code)

include/ui/input-text-model.hpp
```diff
@@ -43,10 +46,16 @@ class InputTextModel : QObject {
...
+   * \param number the number
+   */
+  Q_INVOKABLE void newNumber(std::size_t number);
+
+  /**
    * \brief Sets the new startindex
    * \param text the index
    */
-  Q_INVOKABLE void setStart(unsigned int start);
+  Q_INVOKABLE void setStart(std::size_t start);
```
goldsborough: Since you managed to define a `length_t`, you may just want to rename the `length_t` to `size_t` and use it everywhere. Less typing.

include/ui/input-text-model.hpp
```diff
@@ -66,17 +75,54 @@ class InputTextModel : QObject {
...
 
+  /**
+   * \brief sets the new Mode
+   * \param text the new mode
+   */
+  Q_INVOKABLE void setMode(int mode);
+
+  /**
+   * \brief Gets the mode
+   * \return the length
```
goldsborough, line 94: 😞

include/ui/input-text-model.hpp
```diff
@@ -66,17 +75,54 @@ class InputTextModel : QObject {
...
   Q_INVOKABLE QString getStart();
 
+  /**
+   * \brief sets the new Mode
+   * \param text the new mode
+   */
+  Q_INVOKABLE void setMode(int mode);
+
+  /**
+   * \brief Gets the mode
```
goldsborough: You don't need to use `\brief`, the first line is implicitly `brief` (up to the first dot).
Also, sentences should all end with a dot.

include/ui/input-text-model.hpp
```diff
@@ -66,17 +75,54 @@ class InputTextModel : QObject {
...
+   * \return the length
+   */
+  Q_INVOKABLE int getMode();
+
  private:
+  /**
+   * \brief the context property
```
goldsborough: Write this like this (on one line): `/** The context property. */`

include/ui/input-text-model.hpp
```diff
@@ -66,17 +75,54 @@ class InputTextModel : QObject {
...
+
  signals:
   /**
-   * \brief called when maximum Length is changed
+   * \brief called when maximum Length has changed
    */
   void maximumLengthChanged();
+
+  /**
+   * \brief called when mode has changed
```
goldsborough: First letter capitalized and then a dot.

goldsborough:
I've made some changes to help out. Please fix the documentation, then it's fine.

goldsborough (approved):

janschopohl (approved):

PR #224, opened 2016-12-28
Feature/gui highlight changed register

### LINKED ISSUE #182: Highlight last changed values

riedersa:
closes #182


k-lukas (approved):

source/ui/Components/Register/DefaultRegister.qml
```diff
@@ -95,4 +113,74 @@ TextField {
...
+            height: registerTextField.height
+            color: registerTextField.backgroundColor
+            radius: 2
+            border.color: "lightgray"
+            border.width: 1
+        }
+    }
+
+    //can be used for blinking
+    /*Component{
```
janschopohl, line 132: As this component is not used, I would delete this before merging to master, as this code can not be properly tested by reviewers at the moment. Futhermore, the structure of this seems strange, because it is nested dynamic loading (a component (which has to be created dynamically) loads another component...that should be possible without this indirection). As the highlights are changed quite often, this much dynamic loading might be very inefficient.

janschopohl (approved):
Worked well in my tests

PR #235, opened 2016-12-31
increase memory performance with cache

andizimmerer:


source/ui/memory-component-presenter.cpp
```diff
@@ -100,58 +121,126 @@ MemoryComponentPresenter::data(const QModelIndex &index, int role) const {
...
+    // (done this way to prevent overflows)
+    const int offset = 10;
+    if (_memoryCacheBaseAddress < offset)
+      _memoryCacheBaseAddress = 0;
+    else
+      _memoryCacheBaseAddress -= offset;
+    if (_memoryCacheBaseAddress + _memoryCacheSize >= _memorySize - offset * 2)
+      _memoryCacheSize = _memorySize - _memoryCacheBaseAddress;
+    else
+      _memoryCacheSize += offset * 2;
```
k-lukas, line 225: Don't omit brackets around if-else blocks. To keep the code short, `std::max` and `std::min` could also be used here

source/ui/memory-component-presenter.cpp
```diff
@@ -42,12 +42,33 @@ MemoryComponentPresenter::~MemoryComponentPresenter() {
...
-                   this->index(address + length - 1));// 16bit
-  emit dataChanged(this->index(address / 4),
-                   this->index(address + length - 1));// 32bit
+  // calculate region for (max) 64bit memory cells
+  int start = address - (address % (64 / 8));
+  int end = start + (length - (length % (64 / 8)) + (64 / 8)) - 1;
+
+  // region should not exceed real memory size
+  if (end >= _memorySize) end = _memorySize - 1;
+  if (start <= 0) start = 0;
```
k-lukas: Could be substituted with `std::max`/`std::min` during variable assignment

k-lukas (approved):

include/ui/memory-component-presenter.hpp
```diff
@@ -20,7 +20,7 @@
 #ifndef ERAGPSIM_UI_MEMORY_COMPONENT_PRESENTER_HPP
 #define ERAGPSIM_UI_MEMORY_COMPONENT_PRESENTER_HPP
 
-#include <QAbstractListModel>
+#include <QAbstractTableModel>
```
janschopohl: As you are using the QAbstractListModel again, you should revert this again. :wink:

include/ui/memory-component-presenter.hpp
```diff
@@ -80,8 +80,46 @@ class MemoryComponentPresenter : public QAbstractListModel {
...
+   * /param address Address of the MemoryValue
+   * /param length Size of the MemoryValue
+   * /return The requested MemoryValue
+   */
+  MemoryValue getMemoryValueCached(const std::size_t address, const std::size_t length) const;
+
+  /**
    * Returns the number of rows in this table
-   * Inherited from QAbstractListModel
+   * Inherited from QAbstractTableModel
```
janschopohl: QAbstractListModel

source/ui/memory-component-presenter.cpp
```diff
@@ -42,12 +42,33 @@ MemoryComponentPresenter::~MemoryComponentPresenter() {
...
                                                std::size_t length) {
-  emit dataChanged(this->index(address / 1),
-                   this->index(address + length - 1));//  8bit
-  emit dataChanged(this->index(address / 2),
-                   this->index(address + length - 1));// 16bit
-  emit dataChanged(this->index(address / 4),
-                   this->index(address + length - 1));// 32bit
+  // calculate region for (max) 64bit memory cells
+  int start = address - (address % (64 / 8));
+  int end = start + (length - (length % (64 / 8)) + (64 / 8)) - 1;
```
ProfPille: isn't this the same as `start = address / 8 * 8` and `end = start + (length / 8 * 8) + 7`?

ProfPille: Also: why are you using `int` if your input is `size_t`

source/ui/memory-component-presenter.cpp
ProfPille: `address` is `size_t` and `start` should therefore not be able to be negative (`n>n%m, n>=0`)

source/ui/Components/Memory/MemoryComponent.qml
```diff
@@ -81,17 +84,44 @@ Item {
...
+        // collapse row if cell is not needed
+        Rectangle{
+            height: (styleData.row % (number_bits / 8) == 0) ? 25 : 0
+        }
+    }
+
+    Component {
+        // disable row when cell is not needed
+        id: itemdelegate
+        Text {
```
janschopohl: The QtQuickControls `Label` element might look better on certain systems (You just have to switch to `Label`, everything else should be the same).

source/ui/Components/Memory/MemoryComponent.qml
```diff
@@ -81,17 +84,44 @@ Item {
...
         // when editing is finished the new value is passed to the memory in the core
         id: inputBox
+
         TextField {
             id: textFieldMemoryValue
-            text: styleData.value
+            text: (styleData.row % (number_bits / 8) == 0) ? styleData.value : ""
+            // hide input if cell is not needed
+            visible: (styleData.row % (number_bits / 8) == 0) ? true : false
+            enabled: (styleData.row % (number_bits / 8) == 0) ? true : false
```
janschopohl, line 118: Add `import QtQuick.Controls.Styles 1.4` at the top (if not there already) and add
```
style: TextFieldStyle: {
    renderType: Text.QtRendering
    background: Rectangle {
                    id: styleRec
                    color: "transparent"
    }
}
```
to remove strange rendering artifacts and the textfield outline (The background part is taken from your old memory component).

source/ui/memory-component-presenter.cpp
```diff
@@ -42,12 +42,33 @@ MemoryComponentPresenter::~MemoryComponentPresenter() {
...
+  bool overlap_ending = (address <= _memoryCacheBaseAddress &&
+                         address + length >= _memoryCacheBaseAddress);
+  // new value is completly inside cached region
+  bool overlap_middle =
+      (address >= _memoryCacheBaseAddress &&
+       address + length <= _memoryCacheBaseAddress + _memoryCacheSize);
+
+  if (overlap_beginning || overlap_middle || overlap_ending) {
+    _memoryCacheValid = false;
+  }
```
ProfPille, line 66: couldn't this be written as `start < _memoryCacheBaseAddress + _memoryCacheSize && end > _memoryCacheBaseAddress` ?

source/ui/memory-component-presenter.cpp
```diff
@@ -100,58 +121,126 @@ MemoryComponentPresenter::data(const QModelIndex &index, int role) const {
...
-    // format index as hex value and return it
-    return QString("%1")
-        .arg(index.row() * number_of_bits / 8, 4, 16, QLatin1Char('0'))
-        .toUpper()
-        .prepend("0x");
-  }
 
-  if (role == InfoRole) {
-    // TODO fetch value from core
+  int memory_address = index.row();//* (number_of_bits / 8);
```
janschopohl: remove outdated comments

source/ui/memory-component-presenter.cpp
```diff
@@ -100,58 +121,126 @@ MemoryComponentPresenter::data(const QModelIndex &index, int role) const {
...
+  if (memory_address + memory_length <= _memorySize) {
+    MemoryValue memory_cell =
+        getMemoryValueCached(memory_address, memory_length);
+
+    std::string memoryStringValue;
+    if (role_string.startsWith("bin")) {
+      memoryStringValue = StringConversions::toBinString(memory_cell);
+    }
+    // not yet implemented by conversions
+    // else if (role_string.startsWith("oct"))
```
janschopohl, line 166: I don't think we have to implement octal conversions

source/ui/memory-component-presenter.cpp
```diff
@@ -100,58 +121,126 @@ MemoryComponentPresenter::data(const QModelIndex &index, int role) const {
...
 
-  return QString::fromStdString(stringValue);
+    // calculate new cache region
+    _memoryCacheBaseAddress = memory_address;
+    _memoryCacheSize = memory_length;
+
+    // add some offset around but
+    // cache size should not exceed real memory boundaries
+    // (done this way to prevent overflows)
+    const int offset = 10;
```
ProfPille: You can use `size_t` here. I personally try to avoid int in general as you have no promise of the size, etc. Also mixing signed and unsigned types actually causes a warning ins msvc 😺

source/ui/memory-component-presenter.cpp
```diff
@@ -100,58 +121,126 @@ MemoryComponentPresenter::data(const QModelIndex &index, int role) const {
...
+    _memoryCache =
+        _memoryAccess
+            .getMemoryValueAt(_memoryCacheBaseAddress, _memoryCacheSize)
+            .get();
+    // set cache status to valid
+    _memoryCacheValid = true;
+
+    int subset_start = (memory_address - _memoryCacheBaseAddress) * 8;
+    int subset_end =
+        (memory_address - _memoryCacheBaseAddress + memory_length) * 8;
```
ProfPille, line 238: avoid int

ProfPille (changes requested):

janschopohl (commented):
I think we can remove the `octal` option from the numeric systems.
Otherwise looks quite good, only some minor problems. And the performance is way better, even without the new scheduler! 👍

source/ui/memory-component-presenter.cpp
```diff
@@ -100,58 +121,126 @@ MemoryComponentPresenter::data(const QModelIndex &index, int role) const {
...
-        .arg(index.row() * number_of_bits / 8, 4, 16, QLatin1Char('0'))
-        .toUpper()
-        .prepend("0x");
-  }
 
-  if (role == InfoRole) {
-    // TODO fetch value from core
+  int memory_address = index.row();//* (number_of_bits / 8);
+  int memory_length = number_of_bits / 8;
+
```
janschopohl, line 152: You could add `if(memory_address%memory_length != 0) return QString("")` here as an optimization, as qml strangely keeps calling data in these cases (but nothing has to be displayed).

source/ui/memory-component-presenter.cpp
andizimmerer: `std::max` can not be used in the first assignment because it will compare 0 with a very large number due to an overflow. You are right that `std::min` can be used in the second case. But personally I would like to prefer to keep both cases similar.

source/ui/memory-component-presenter.cpp
andizimmerer: I would like to keep the octal conversions commented out. It thinks it's quite likely that someone could implement this feature in the future.

source/ui/memory-component-presenter.cpp
andizimmerer: Yes, absolutely true! :)
But if you don't mind I would leave it this way. It doesn't make any difference :D

ProfPille (approved):

PR #240, opened 2017-01-04
Feature/swap store operand order

### LINKED ISSUE #234: Base and source registers in store instructions should be swapped

goldsborough:
closes #234


include/arch/riscv/load-store-instructions.hpp
```diff
@@ -355,14 +230,7 @@ class StoreInstructionNode
...
 
-  MemoryValue getBase(MemoryAccess& memoryAccess) const override {
-    return super::_children.at(0)->getValue(memoryAccess);
-  }
-
-  MemoryValue getOffset(MemoryAccess& memoryAccess) const override {
-    return super::_children.at(2)->getValue(memoryAccess);
-  }
-
+  /** The type of store instruction this is. */
```
ErikP0, line 233: ~this is~

ErikP0 (approved):
Please also change the order of operands in the help messages (see `instruction-context-information.cpp` line 466)

janschopohl (approved):
Works as expected and the code looks good

PR #246, opened 2017-01-06
add 64bit view in memory; memory styling

andizimmerer:


source/ui/Components/Memory/MemoryComponent.qml
```diff
@@ -109,26 +109,64 @@ Item {
...
-                                id: styleRec
-                                color: "transparent"
+        MouseArea {
+            hoverEnabled: true
+            anchors.fill: parent
+            // fadein on mouse hover
+            onHoveredChanged: {
+                if(containsMouse) {
+                    textFieldMemoryValue.borderopacity = 1;
+                } else {
```
goldsborough: `else if (...)`

source/ui/Components/Memory/MemoryComponent.qml
```diff
@@ -109,26 +109,64 @@ Item {
...
-            onEditingFinished: {
-                // update internal memory; use right number representation and byte size
-                // if cell is not needed we can save an update
-                if(styleData.row % (number_bits / 8) == 0) {
-                    memoryModel.setValue(styleData.row, textFieldMemoryValue.text, number_bits, tableView.getColumn(styleData.column).role);
+            TextField {
+                id: textFieldMemoryValue
+                text: (styleData.row % (number_bits / 8) == 0) ? styleData.value : ""
+                anchors.fill: parent
+                visible: (styleData.row % (number_bits / 8) == 0) ? true : false
```
goldsborough, line 128: s/`(condition) ? true : false`/`condition`

source/ui/Components/Memory/MemoryComponent.qml
```diff
@@ -109,26 +109,64 @@ Item {
...
+                enabled: (styleData.row % (number_bits / 8) == 0) ? true : false
+
+                onActiveFocusChanged: {
+                    if(activeFocus) {
```
goldsborough: Can use the ternary operator here `?:`

goldsborough (approved):

ErikP0 (approved):
Hi,
As far as I can see, your code looks good!
I tried out the memory component and found some suggestions you may change:
- I was able to remove the last column showing the memory values (not the addresses). See picture attached. I suggest to remove the "remove.." option on the last displayed column, because a memory component showing only the addresses is meaningless.
- I suggest to change the text for the unsigned decimal displaying option from "decimal" to "decimal (unsigned)" in order to be more consistent. The register components display options are labeled that way.
- Is it possible to change the default number representation from bit to hex, as the bit representation is too long (never fully displayed in the default sized text field)?
- When I create a project with less than 20 bytes memory, the program crashes.

ErikP0:
memory_comp

PR #248, opened 2017-01-09
Feature/ui stylesheets

goldsborough:
This PR provides the initial architecture for styling the GUI, along with some initial styling, mainly of the creation screen.

## General Overview

Styling is implemented as follows:

1. Themes are to be stored in `.erasim/themes/` as folders with a `.theme` extension. Later `.erasim` will be in the user's `$HOME` directory, for now it is in the root folder of the project. A theme consists of three files (more about these files below):
    1. A SASS file (`theme.sass`), which should be the primary working document.
    2. A CSS file (`theme.css`), generated from the SASS file by the SASS compiler.
    3. A JSON file (`theme.json`), generated from the CSS file by `theme2json.py`.
2. Once the JSON file was generated, it is loaded into the QT application at runtime and becomes accessible in QML via the `Theme` Item. The SASS/CSS selectors map directly to JSON properties, such that they can be used easily. Because the `Theme` is a `QmlPropertyMap`, the styling is done via dynamic bindings. This means that the application need not be recompiled to change styles. In fact, the stylesheet can be changed at runtime!

## SASS -> CSS

We develop our stylesheets using [SASS](http://sass-lang.com), a CSS preprocessor. Basically, SASS is like CSS, with a more comfortable, indentation-based syntax and allowing the use of variables, functions and generally more complex programming. While you would write the following in CSS:

```css
.a, .b {
 width: 50px;
 color: red;
}

.b {
  margin: 55px;
}
```

You can now write:

```sass
$width: 50px

.a
  width: $width
  color: red

.b
  @extends .a
  margin: $width + 5px
```

To go from SASS to CSS, one then requires the `sass` compiler command-line utility. It can be downloaded as a Ruby gem on any popular platform. See [here](http://sass-lang.com/install) for instructions. Then, you'd usually want to set up a persistent watch on your SASS file, e.g.:

```sh
$ sass --watch theme.sass:theme.css
```

This will automatically compile your SASS theme to CSS whenever you make any change to the SASS file.

## CSS -> JSON

The simulator is not able to understand CSS per se. We first need to transform the CSS into a format that we can load and access in our application. For this, we convert the CSS file to JSON. This is done using my script `theme2json.py`. In its basic functionality, it takes as input a CSS file and converts it to JSON:

```sh
$ pwd
/path/to/era-gp-sim/
$ python3 scripts/theme2json.py .erasim/themes/my_theme.theme/theme.css
```

This will, by default, emit a `theme.json` file under the same directory as the input file. For convenience, `theme2json.py` also accepts the path to the SASS file and will invoke the SASS compiler for you (it still needs to be installed). For even greater convenience, you can also simply pass the path of the theme folder (ending in `.theme`). This will convert the SASS file to CSS and then the CSS to JSON. For more information on the script, pass the `-h` option:

```sh
usage: theme2json.py [-h] [-v] [-o FILE] [-c | -s | -d] INPUT

Theme to JSON converter.

positional arguments:
  INPUT                 The theme folder, SASS or CSS to convert.

optional arguments:
  -h, --help            show this help message and exit
  -v, --verbose         Give more information about the program's proceedings
  -o FILE, --output FILE
                        Stores the generated JSON to FILE
  -c, --css             INPUT is a CSS file.
  -s, --sass            INPUT is a SASS file.
  -d, --directory       INPUT is a directory.
```

You can control the verbosity of the output with the `-v` option. Passing `-vv` makes it extra verbose.

## Naming Transformations

How does a SASS (or CSS) file map to JSON? Basically, each level of selector nesting in CSS maps to one level of object nesting in JSON, with any special characters in names replaced by `camelCasing`. For example, this SASS file:

```sass
my-button
  color: red
  border
    width: 5px
    style: dotted
```

which is equivalent to the following CSS:

```css
my-button {
  color: red;
}

my-button border {
  width: 5px;
  style: dotted;
}
```

will be converted to the following JSON (note the conversion from `hy-phe-nated` to `camelCase`):

```json
{
  "myButton": {
    "color": "red",
    "border": {
      "width": 5,
      "style": "dotted"
    }
  }
}
```

and can then be accessed like so in QML:

```qml
Rectangle {
  color: Theme.myButton.color
  border {
    width: Theme.myButton.border.width
  }
}
```

Note that classes (starting with a dot), IDs (starting with a hash) and what would normally be HTML elements (e.g. `div`, `button` etc.) are treated all the same and result in one level of nesting. This is because JSON has no concept of specificity as CSS would have (where IDs have more weight than classes, for example).

One more thing to note is that during conversion from CSS to JSON, *all absolute units are thrown away*. That is, `5px` is the same as `5rem` or `5em` or `5vw` to `theme2json.py`. Also, percentage units are converted to a scalar value in `[0, 1]`, i.e. `50%` will map to `0.5` so it can be used in QML directly. More precisely, the conversion rules are given below:

Let `p` be a property found in the CSS file. Match `p` against the regular expression `^(?:([+-]?\d*\.\d+)|([+-]?\d+))\s*(%)?`. Then:

1. If `p` does not match, convert `p` to a string. This means any non-numeric properties are always stringified.
2. Else, if `p` matches the first group (`[+-]?\d*\.\d+`), it is interpreted as a float (e.g. `0.5`, `.3`).
3. Else and given (1) is not the case (the string matches the regular expression), `p` must be an integer (matching `[+-]?\d+`) and is interpreted as such.

If, after (2) or (3) occurred, also the third group (`%`) matches, `p` is interpreted as a percentage and its value divided by 100.

This logic can be seen live by passing `-vv` when converting a CSS file with `theme2json.py`.


source/ui/ui.cpp
```diff
@@ -17,10 +17,11 @@
...
 */
 
-#include "ui/ui.hpp"
-#include "ui/clipboard-adapter.hpp"
-
 #include <QUrl>
+#include <string>
+
+#include "ui/clipboard-adapter.hpp"
+#include "ui/ui.hpp"
```
janschopohl, line 24: ui.hpp should still be above the library includes, right? I don't know who added clipboard-adapter.hpp at the top there.

It also seems like the order of the includes is not alphabetical now (ui/..., arch/..., ..., ui/...)

source/ui/theme.cpp
```diff
@@ -0,0 +1,102 @@
...
+  assert::that(_theme != nullptr);
+  return *_theme;
+}
+
+Theme::Theme(const QString& themeName, QObject* parent) : super(this, parent) {
+  assert::that(!themeName.isEmpty());
+  load(themeName);
+}
+
+void Theme::load(const QString& themeName) {
```
janschopohl, line 49: Maybe we should use exceptions here, as the user could load an invalid theme. You could catch the exception in the configuration code and display an error in that case.

I guess it depends on how you want to handle the theme configuration.

source/ui/main.qml
```diff
@@ -24,240 +24,252 @@ import QtQuick.Dialogs 1.2
...
-            // create a new tab if there is no tab anymore, prevents a blank screen.
-            createProject();
+  function showMenus() {
+    toolbar.showToolbar();
+    menubar.setMenuEnabled(true);
+  }
+
+  function expand() {
+    window.width = Theme.window.width;
+    window.height = Theme.window.height;
```
janschopohl: Specifying a percentage of the screen size in the theme might be more portable.

source/ui/foo.cpp
```diff
@@ -0,0 +1,20 @@
...
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "ui/foo.hpp"
```
janschopohl, line 20: ?

source/ui/ProjectCreationScreen.qml
```diff
@@ -37,156 +45,212 @@ Item {
...
+          right: parent.right
+          rightMargin: Theme.createProject.section.margin
+          verticalCenter: parent.verticalCenter
         }
-        model: ui.getArchitectures();
-        onCurrentIndexChanged: {
-          var currentArchitecture = model[currentIndex];
-          optionSelector.model = ui.getOptionNames(currentArchitecture);
-          parserSelector.model = ui.getParsers(currentArchitecture);
+        style: ComboBoxStyle {
```
janschopohl, line 172: In case you didn't notice it on your system:
When adding a `ComboBoxStyle`, the default system styling is overriden by default Qt styling. The difference can be seen here:
screenshot_20170111_141513 
(I removed the `ComboBoxStyle` from the lowest ComboBox for this screenshot)
The best solution is probably to completely style the `ComboBoxStyle` with the theme (custom background,...).
You could create a `StyledComboBox` qml element, as the element is used by multiple other components.

scripts/theme2json.py
```diff
@@ -0,0 +1,658 @@
...
+            log.debug('Yielding selector leaf %s', selector)
+            yield parent
+
+
+def process_property_value(value):
+    """
+    Processes the value of a property.
+
+    We do not respect any kind of unit for numbers. For this script (and our
+    themes), 10px is the same as 10% and 10rem or 10em. In fact, the unit is
```
janschopohl: well 10% is processed to 0.1, so I guess it's not the same as 10px :wink:

.erasim/themes/solarized.theme/theme.sass
```diff
@@ -0,0 +1,47 @@
...
+  border
+    color: lighten($primary-color, 20%)
+    width: 2px
+
+window
+  width: 1124px
+  height: 700px
+  background-color: white
+  initial-width: $section-width + 20px
+  initial-height: $section-height * 6 + $button-height + 10
```
janschopohl, line 59: when styling the tabbar, its height should be added here as well to adjust the initial height properly on all systems.

source/ui/ProjectCreationScreen.qml
```diff
@@ -37,156 +45,212 @@ Item {
...
-    //select a formula for the architecture
-    ComboBox {
-        id: optionSelector
-        anchors {
-          top: architectureSelector.bottom
-          left: parent.left
-          right: parent.horizontalCenter
+      MouseArea {
+        anchors.fill: button
+        cursorShape: Qt.PointingHandCursor
```
janschopohl, line 238: Do you use the MouseArea inside the button only for the cursorShape? I think it would be better to create a custom Button in that case to avoid overlaying an additional mouse area. It would also be great for reusability, as there is quite much styling in this button.

source/ui/ProjectCreationScreen.qml
```diff
@@ -37,156 +45,212 @@ Item {
...
-        }
-        text: "create Project"
-        onClicked: {
-            tab.title = textInputName.text;
-            root.buttonClicked(memorySizeSelector.getCurrentValue(),
-              architectureSelector.currentText, optionSelector.currentText, parserSelector.currentText);
+    ProjectCreationSection {
+      id: projectName
+      anchors.topMargin: Theme.createProject.marginTop
+      TextField {
```
janschopohl, line 55: If you want to avoid an empty project name, you could add something like this:
```
onTextChanged: {
  if (text === "" || !acceptableInput) {
    button.enabled = false;
  } else {
    button.enabled = true;
  }
}
```
and in the `Button`:
```
enabled: false
```
to initialize it properly.

janschopohl (approved):
I only found some minor things, seems to work quite well.
I also quickly looked over the python code and didn't notice any major issue.

scripts/theme2json.py
goldsborough: true thanks

source/ui/ProjectCreationScreen.qml
goldsborough: Ok

source/ui/foo.cpp
goldsborough: lel

source/ui/theme.cpp
goldsborough: There will be a fixed number of themes that the user can select from a combo box, so unless the user messes with the configuration, nothing bad should happen. But for that case, I'll add some handling. Thanks.

source/ui/ProjectCreationScreen.qml
```diff
@@ -37,156 +45,212 @@ Item {
...
-        anchors {
-          top: architectureSelector.bottom
-          left: parent.left
-          right: parent.horizontalCenter
+      MouseArea {
+        anchors.fill: button
+        cursorShape: Qt.PointingHandCursor
+        onClicked: {
+          root.createProject(
+            projectName.text,
```
janschopohl: this should be `projectNameField.text`. Doesn't work currently.

PR #252, opened 2017-01-11
Fix for #251

### LINKED ISSUE #251: Memory does not update properly
The memory does not update properly on reset currently. I think i found the error here:
https://github.com/TUM-LRR/era-gp-sim/blob/8eac5fd44e9ab130d367b45acc5832fe0df9a1d2/source/ui/memory-component-presenter.cpp#L48
`std::max` should propably be `std::min`. When changing this, everything should work again.

janschopohl:
closes #251


FelixOpolka (approved):
Everything works as expected on my computer.

PR #253, opened 2017-01-11
Fix crash in parser on division by zero

### LINKED ISSUE #249: Arithmetic exceptions crash the simulator
When an arithmetic exception is generated while calculating an immediate, the simulator crashes:
Ex: `addi x0, x0, 1/0`

kreudom:
closes #249
This fix introduces the following changes:

* Check if divisor is 0 for the `/` and `%` operators and record an error in that case
* A CodePositionInterval parameter was added to the operator definitions to determine error positions
* Added a test case for `1/0`


davschneller (approved):
Yes, this is great like that. Even better would be maybe to even mark the whole sub expression with the error (not just the current token), i.e. got on the output stack not only the IntType, but also a CodePositionInterval which grows bigger as more numbers/expressions it swallows. This might offer an even more accurate view on the errors. But it is totally ok like it is right now, as we only look at the lines in error displaying. 👍 (and sorry for the review delay)

include/parser/independent/expression-compiler-clike.hpp
```diff
@@ -41,202 +50,246 @@ namespace CLikeExpressionCompilers {
...
-                                           IntType& out,
-                                           CompileErrorList& errors) -> bool {
-                                          out = f % s;
-                                          return true;
-                                        }}};
+      ExpressionBinaryOperator<IntType>{
+          ExpressionOperatorAssociativity::LEFT,
+          "||",
+          120,
+          [](const IntType& f,
```
goldsborough: Please use better variable names, I have no idea what `f` or `s` are from reading this

include/parser/independent/expression-compiler-clike.hpp
```diff
@@ -28,6 +28,15 @@
 #include "parser/independent/integer-parser.hpp"
 #include "parser/independent/string-parser.hpp"
 
+#define clike_recordError(errorList, position, msg, ...) \
```
goldsborough: Why is this a macro? And if there is a good reason, it should be ALL_CAPS

include/parser/independent/expression-compiler-clike.hpp
```diff
@@ -41,202 +50,246 @@ namespace CLikeExpressionCompilers {
...
-                                        }}};
+      ExpressionBinaryOperator<IntType>{
+          ExpressionOperatorAssociativity::LEFT,
+          "||",
+          120,
+          [](const IntType& f,
+             const IntType& s,
+             IntType& out,
+             CompileErrorList& errors,
+             const CodePositionInterval& pos) -> bool {
```
goldsborough: You don't need trailing return types (`bool` can be deduced from `return true`)

include/parser/independent/expression-compiler-clike.hpp
```diff
@@ -41,202 +50,246 @@ namespace CLikeExpressionCompilers {
 template <typename IntType>
 ExpressionCompiler<IntType> createCLikeCompiler() {
   auto binaryOperators = std::vector<ExpressionBinaryOperator<IntType>>{
```
goldsborough: It never makes sense to do `auto x = Type(...)`. That's not what `auto` is good for. Just write `Type x(...)`

goldsborough: Or in this case `Type x = {...}`

include/parser/independent/expression-compiler-clike.hpp
```diff
@@ -41,202 +50,246 @@ namespace CLikeExpressionCompilers {
...
+          120,
+          [](const IntType& f,
+             const IntType& s,
+             IntType& out,
+             CompileErrorList& errors,
+             const CodePositionInterval& pos) -> bool {
+            out = f || s ? IntType(1) : IntType(0);
+            return true;
+          }},
+      ExpressionBinaryOperator<IntType>{
```
goldsborough: Why don't you define a helper function and reduce this file by 80000000000%?

```
template<typename IntType, typename Implementation>
ExpressionBinaryOperator<IntType> binaryOperator(const char* token, std::size_t, someNumber, Implementation implementation) {
  return {
    token, 
    someNumber,
    ExpressionOperatorAssociativity::LEFT,
     [] (const IntType& f,
          const IntType& s,
           IntType& out,
           CompileErrorList& errors,
           const CodePositionInterval& pos) -> bool {
            out = implementation(f, s);
            return true;
          }},
}

...

// For example
binaryOperator<IntType>("||", 120, [](auto& first, auto& second) { return first || second; });
```

include/parser/independent/expression-compiler-clike.hpp
```diff
@@ -41,202 +50,246 @@ namespace CLikeExpressionCompilers {
...
+          ExpressionOperatorAssociativity::LEFT,
+          "&&",
+          110,
+          [](const IntType& f,
+             const IntType& s,
+             IntType& out,
+             CompileErrorList& errors,
+             const CodePositionInterval& pos) -> bool {
+            out = f && s ? IntType(1) : IntType(0);
```
goldsborough: These conversions are useless and should be `static_cast` anyway

goldsborough (approved):

include/parser/independent/expression-compiler-clike.hpp
ErikP0: Due to the translation concept of Qt consisting mostly of macros (e.g. `QT_TRANSLATE_NOOP`) we use macros to simplify the expression around the message to be marked for translation. So in my opinion this is fine.
But it should be ALL_CAPS.

ErikP0 (approved):
The fix looks good, altough using @goldsborough 's suggestion for this helper function would really enhance readability.

PR #254, opened 2017-01-14
Feature/ui stylesheets-enhancements

### LINKED ISSUE #159: ui sylesheets

goldsborough:
closes #159 
This PR brings two main changes to the UI theming code (or changes *for* it):

1. Refactoring of the `ProjectCreationScreen` into one file per section.
2. Creation of a `Status` and `StatusWithValue` class to propagate "exceptions" in Qt without needing to compile the Qt binary with exceptions. This is needed to catch very early errors that would prevent starting the UI because of some basic problems, such as the theme (or, later, configuration) folder not being found. By "throwing" an exception (returning error codes) we can catch such problems before loading the main QML file and instead loading only an error dialog showing the problem.

You can try it out by making a spelling mistake in `ui.cpp`, line 57:

All OK:
```cpp
auto status = Theme::Make("default");
```

Typo causes error box to show:
```cpp
auto status = Theme::Make("banana"); // minor typo
```


include/common/status.hpp
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * \returns The message of the status.
+   */
+  const std::string& message() const noexcept;
+
+  /**
+   * \returns True if the status code is OK, else false.
+   */
+  explicit operator bool() const noexcept {
+    return _code == Code::OK;
```
janschopohl: -> cpp

include/common/status-with-value.hpp
```diff
@@ -0,0 +1,126 @@
...
+  const T value() const noexcept {
+    assert::that(static_cast<bool>(_value));
+    return *_value;
+  }
+
+  /**
+   * \returns The status.
+   */
+  const Status& status() const noexcept {
+    return _status;
```
janschopohl, line 90: I guess this should be in the cpp file, for better style.

include/ui/theme.hpp
```diff
@@ -52,24 +54,41 @@ class Theme : public QQmlPropertyMap {
...
+   *
+   * This factory function should only be called once.
+   * The class does not take ownership of the allocated theme, it
+   * only provides access to it.
+   *
+   * \param themeName The theme to initialize the new theme singleton with.
    * \returns A pointer to the newly created theme.
    */
-  static Theme* reset(const QString& themeName, QObject* parent = nullptr);
+  static Status Make(const QString& themeName);
```
janschopohl, line 66: Is there a reason for uppercase?

janschopohl (approved):
I tested this quite a bit and everything worked for me.
👍

include/ui/theme.hpp
goldsborough: It's a factory function (see also `NodeFactoryCollection`)

include/common/status-with-value.hpp
goldsborough: Unfortunately it's a template :(

PR #256, opened 2017-01-15
Feature/ui configuration

goldsborough:
This PR brings the basic configuration (henceforth referred to as *settings*) infrastructure to the simulator. There are three main contributions:

1. The addition of a settings window, available as an additional menu (or under `Ctrl+,`).
2. The implementation of a `Settings` object that can load and store settings from the window.
3. Ability to configure the snapshot location as well as theme.

A nice feature is that the theme can be loaded "online", i.e. changes take effect instantly.


.erasim/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "snapshotLocation": "/Users/goldsborough/Documents",
```
janschopohl: You want to force a goldsborough user on everyone, don't you? 😜

include/ui/gui-project.hpp
```diff
@@ -290,21 +356,21 @@ class GuiProject : QObject {
...
-   * \brief The Functions for the conversion
+   * The Functions for the conversion
    */
-  std::function<std::string(MemoryValue)> hexConversion;
-  std::function<std::string(MemoryValue)> binConversion;
-  std::function<std::string(MemoryValue)> octConversion;
-  std::function<std::string(MemoryValue)> signedDecimalConversion;
-  std::function<std::string(MemoryValue)> unsignedDecimalConversion;
-  std::function<std::string(MemoryValue)> decimalFloatConversion;
+  MemoryToStringConverter hexConversion;
```
janschopohl: I guess you could add `_` to these names if you are already improving the style.

source/ui/Components/Menubar/Menubar.qml
```diff
@@ -22,111 +22,127 @@ import QtQuick 2.0
...
-            text: "Save Snapshot"
-            onTriggered: {
-                actionSnapshot();
-            }
-        }
+    MenuItem {
+      text: "Save Snapshot As"
+      shortcut: "Ctrl+Shift+S"
+      onTriggered: {
+        actionSnapshot();
```
janschopohl: Does the same as Save Snapshot currently, right?

source/ui/Components/ProjectCreation/ProjectCreationScreen.qml
```diff
@@ -17,7 +17,7 @@
 * along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import QtQuick 2.6
+import QtQuick 2.7
```
janschopohl: No, we need to use qt 5.6. Use `import QtQuick 2.6` instead

source/ui/Components/Snapshots/snapshotmodel.cpp
```diff
@@ -17,80 +17,75 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
```
janschopohl: Wow, I was really confused what this file is, but I realized that it is not used anywhere. Can be deleted.

source/ui/Components/Settings/SettingsWindow.qml
```diff
@@ -0,0 +1,110 @@
...
+      store();
+
+      Settings.theme = themes.selection;
+      Settings.snapshotLocation = snapshotLocation.location;
+
+      var message = Settings.store();
+      if (message.length > 0) {
+        errorDialog.show(message);
+      } else {
+        hasUnsavedChanges = false;
```
janschopohl: Something doesn't work here: If I change to the other theme and save, and then try to change back to default, I can't save my changes. Seems like `hasUnsavedChanges` is not set properly in some cases.

source/ui/Components/Menubar/Menubar.qml
```diff
@@ -22,111 +22,127 @@ import QtQuick 2.0
...
+    main.fileDialog.onAcceptedFunction = saveAs;
+    main.fileDialog.selectExisting = false;
+    main.fileDialog.open();
+  }
+
+  function setMenuEnabled(value) {
+    openFileOption.enabled = value;
+    saveFileOption.enabled = value;
+    saveFileAsOption.enabled = value;
+    projectMenu.enabled = value;
```
janschopohl: Minor problem: You can't close a project creation screen tab anymore.

janschopohl (changes requested):
Some minor things broke, otherwise works really good!

Snapshots are dependent on your name again at the moment. :wink:

source/ui/Components/Menubar/Menubar.qml
goldsborough: Yes, getting rid of

PR #270, opened 2017-01-18
Feature/theme snapshots

goldsborough:
Theming for se snapshots


source/ui/Components/Snapshots/AddSnapshotButton.qml
```diff
@@ -0,0 +1,30 @@
...
+* You should have received a copy of the GNU General Public License
+* along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import QtQuick 2.0
+import QtQuick.Controls 1.4
+import QtQuick.Controls.Styles 1.4
+
+import Theme 1.0
+
```
janschopohl: I think this file is unused?

source/ui/Components/Snapshots/Divider.qml
```diff
@@ -0,0 +1,33 @@
...
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import QtQuick 2.0
```
janschopohl: 2.6

source/ui/Components/Snapshots/SnapshotDialog.qml
```diff
@@ -0,0 +1,97 @@
...
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import QtQuick 2.0
```
janschopohl: 2.6

source/ui/Components/Snapshots/SnapshotHeader.qml
```diff
@@ -0,0 +1,56 @@
...
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import QtQuick 2.0
```
janschopohl: 2.6

source/ui/Components/Snapshots/SnapshotItem.qml
```diff
@@ -0,0 +1,63 @@
...
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import QtQuick 2.0
```
janschopohl: 2.6

source/ui/Components/Snapshots/SnapshotList.qml
```diff
@@ -1,187 +1,57 @@
...
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
 
 import QtQuick 2.0
```
janschopohl: 2.6

source/ui/Components/Snapshots/SnapshotItem.qml
```diff
@@ -0,0 +1,63 @@
...
+import QtQuick.Controls.Styles 1.4
+
+import Theme 1.0
+
+Component {
+  id: listDelegate
+  Item {
+    id: item
+    property alias text: snapshot.text
+    property bool active: listView.currentIndex == model.index
```
janschopohl, line 33: I dont't know if this is intended, but a snashot entry stays highlighted, even if the snapshot window lost focus. Maybe you have to add a onFocusChanged: currentIndex = -1 in the snapshot list, or something else to stop highlighting.

janschopohl (approved):
The styling looks good, but a change while refactoring (https://github.com/TUM-LRR/era-gp-sim/commit/e1eb932e9e692ae4803164a8cb47ce3fac49fb24) causes the following problem (I think thats the cause):
If you want to oberwrite a snapshot by saving to the same name again, the snapshot is displayed twice in the snapshot list. This was not the case before. But as this is not directly related to this pr, I will create a separate issue for that.

source/ui/Components/Snapshots/SnapshotItem.qml
goldsborough: Isn't it better to always know what the current snapshot is?

source/ui/Components/Snapshots/AddSnapshotButton.qml
goldsborough: yep

source/ui/Components/Snapshots/SnapshotItem.qml
janschopohl: Well I just think it's a bit strange, as after selecting a snapshot the memory and registers can change while the snapshot stays highlighted, so I feel like this doesn't really add anything and looks slightly confusing.

source/ui/Components/Common/ThemeableCheckBox.qml
```diff
@@ -0,0 +1,75 @@
...
+    anchors.fill: parent
+    cursorShape: Qt.PointingHandCursor
+    onClicked: checked = !checked
+  }
+
+  // Returns a style in the Theme.checkbox or Theme.checkbox.active object
+  // depending on whether this checkbox is active. For example,
+  // dynamicStyle('foo.bar') will return Theme.checkbox.active.foo.bar when
+  // the checkbox is active, and Theme.checkbox.foo.bar otherwise.
+  function dynamicStyle(children) {
```
janschopohl: Shouldn't you use the version in ThemeUtilities.js?

source/ui/Components/Settings/Setting.qml
```diff
@@ -20,55 +20,75 @@ import QtQuick 2.6
...
-  property var bottomAnchor
+  property bool isSetting: true // A 'tag' to identify a Setting instance
+
+  property var value
+  property string setting
+  property var widget
+
+  property bool hasChanged: (value !== undefined) && Settings[setting] !== value
+
+  function dynamicStyle(children) {
```
janschopohl: ThemeUtilities.js
?

source/ui/Components/Settings/SettingsWindow.qml
```diff
@@ -87,25 +63,69 @@ Window {
...
+      event.accepted = false;
+    } else {
+      internalCloseOverride = false;
+    }
+  }
 
-  Button {
-    text: "Save"
-    enabled: hasUnsavedChanges
+  ScrollView {
```
janschopohl, line 81: I think using a `ScrollViewStyle` with `transientScrollBars: true` would be nice.
Currently the vertical scrollbar reduces the content width, sometimes causing a horizontal scrollbar to appear. Transient scrollbars are only displayed when used and go over the content (used in help tooltips and registers)

source/ui/Components/Settings/default-settings.json
```diff
@@ -1,4 +1,6 @@
-{
+{
     "snapshotLocation": null,
-    "theme": "default"
+    "theme": "default",
+    "removeSnapshotsPermanently": null,
```
janschopohl, line 4: I think `removeSnapshotsPermanently` and `overrideSnapshots` should default to `false` instead of `null`, otherwise you have unsaved settings at the beginning.

source/ui/Components/Snapshots/SnapshotDialog.qml
```diff
@@ -0,0 +1,99 @@
...
+      enabled: name.text.length > 0
+      text: 'Save'
+      isDefault: true
+      anchors {
+        top: cancel.top
+        bottom: cancel.bottom
+        right: cancel.left
+      }
+      onClicked: {
+        ui.saveSnapshot(tabView.getCurrentProjectId(), name.text);
```
janschopohl: Shouldn't you check here if a snapshot with the same name already exists and display a warning dialog (if not disabled in the settings)? That doesn't work for me currently.
You can directly access the snapshot component through the `snapshotComponent` context property to do this.

source/ui/Components/Snapshots/SnapshotItem.qml
```diff
@@ -0,0 +1,81 @@
...
+      onDone: removeSnapshot(answerWasYes)
+    }
+
+    TextButton {
+      id: removeButton
+      anchors.bottom: divider.bottom
+      text: "\u00D7"
+      theme: Theme.snapshots.item.remove
+      onClicked: {
+        if (removalDialog.alreadyAsked) {
```
janschopohl, line 67: There is supposed to be a dialog here which asks if you want to remove the snapshot permanently, right?
It somehow doesn't work for me...

source/ui/Components/Toolbar/ToolbarMainWindow.qml
```diff
@@ -0,0 +1,202 @@
...
+/*
+  This file implements the toolbar for the Main Window
+ */
+
+import QtQuick 2.6
+import QtQuick.Controls 1.4
+import QtQuick.Layouts 1.3
+import QtQuick.Controls.Styles 1.4
+
+ToolBar {
```
janschopohl: What's going on here? Wasn't this changed and reviewed on another branch (ToolbarComponent.qmll)?

source/ui/snapshot-component.cpp
```diff
@@ -49,17 +49,30 @@ QStringList SnapshotComponent::getSnapshotList(const QString& architecture) {
...
+  }
 }
 
 void SnapshotComponent::removeSnapshot(const QString& architectureIdentifier,
-                                       const QString& snapshotName) {
-  auto path = snapshotPath(architectureIdentifier, snapshotName);
-  QFile(QString::fromStdString(path)).remove();
+                                       const QString& snapshotName,
+                                       bool removePermanently) {
+  if (removePermanently) {
```
janschopohl, line 68: So if removePermanently is false, the snapshot is just deleted from the list and not displayed, but loaded again once the simulator is restarted? Is this intended behaviour?

janschopohl (approved):
I'm not quite sure if all features work as intended, but otherwise it looks good!
:+1:

source/ui/snapshot-component.cpp
goldsborough: Yes, that is the point

source/ui/Components/Snapshots/SnapshotItem.qml
goldsborough: Indeed. For future reference: `null` in JSON translates to an empty `QVariant`, which translates to `undefined` in QML. Why? For the glory of Satan of course!

source/ui/Components/Settings/default-settings.json
goldsborough: I have fixed this by requiring at least one visible change to a setting before it is marked as having changed (such that even if the initial value is different from the setting, because in this case the setting must be a boolean but is really `undefined`/`null`, this is not marked as a change)

source/ui/Components/Settings/SettingsWindow.qml
goldsborough: Fixed

PR #286, opened 2017-01-24
Better support for builtin macros

kreudom:
Errors while parsing built-in macros are now placed at the position of the macro call.

Additionally, executeNextLine will now execute Nodes until a valid line is reached again. This means built-in macros are executed in one step.

### DIFF
--- a/source/core/parsing-and-execution-unit.cpp
+++ b/source/core/parsing-and-execution-unit.cpp
@@ -84,7 +84,7 @@ void ParsingAndExecutionUnit::executeNextLine() {
 
     nextNode = _updateLineNumber(nextNode);
 
-    if (nextNode >= _filenalRepresentation.commandList().size()) break;
+    if (nextNode >= _finalRepresentation.commandList().size()) break;
 
     // Skip nodes that aren't part of the users program
   } while (_finalRepresentation.commandList()[nextNode].position().isEmpty());
--- a/source/core/parsing-and-execution-unit.cpp
+++ b/source/core/parsing-and-execution-unit.cpp
@@ -83,6 +83,9 @@ void ParsingAndExecutionUnit::executeNextLine() {
     if (!_executeNode(nextNode)) break;
 
     nextNode = _updateLineNumber(nextNode);
+
+    if (nextNode >= _filenalRepresentation.commandList().size()) break;
+
     // Skip nodes that aren't part of the users program
   } while (_finalRepresentation.commandList()[nextNode].position().isEmpty());
   _executionStopped();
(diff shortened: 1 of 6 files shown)


source/parser/independent/intermediate-macro-instruction.cpp
```diff
@@ -98,8 +99,36 @@ void IntermediateMacroInstruction::execute(
...
-    operation->execute(immutable, errors, commandOutput, memoryAccess);
+    operation->execute(immutable, innerErrors, innerCommands, memoryAccess);
+  }
+  // If the inner commands don't have their own position, use the position of
+  // the macro call
+  if (innerCommands.size() > 0 && innerCommands[0].position().isEmpty()) {
+    commandOutput.emplace_back(innerCommands[0].node(),
+                               positionInterval(),
+                               innerCommands[0].address());
+    for (auto& error : innerErrors.errors()) {
```
goldsborough: `const auto&`

goldsborough (approved):

source/core/parsing-and-execution-unit.cpp
```diff
@@ -79,9 +79,15 @@ void ParsingAndExecutionUnit::execute() {
...
-    _updateLineNumber(nextNode);
-  }
+  do {
+    if (!_executeNode(nextNode)) break;
+
+    nextNode = _updateLineNumber(nextNode);
+
+    if (nextNode >= _finalRepresentation.commandList().size()) break;
+
+    // Skip nodes that aren't part of the users program
```
davschneller: `user's`

source/parser/independent/intermediate-macro-instruction.cpp
```diff
@@ -98,8 +99,36 @@ void IntermediateMacroInstruction::execute(
     CompileErrorList& errors,
     FinalCommandVector& commandOutput,
     MemoryAccess& memoryAccess) {
+  CompileErrorList innerErrors;
+  FinalCommandVector innerCommands;
   for (const auto& operation : _operations) {
-    operation->execute(immutable, errors, commandOutput, memoryAccess);
+    operation->execute(immutable, innerErrors, innerCommands, memoryAccess);
+  }
+  // If the inner commands don't have their own position, use the position of
```
davschneller, line 107: Do we really only want to do this for the build-in macros? (as far as I could read that out of your code)
What, if the user designs a small macro, but then uses it with wrong syntax like 100 times? Wouldn't that cause a huge bunch of error messages in the macro definition? I'm rather a fan of showing it at the macro call position consequently of each call. Any objections to this?
I mean, the best would be to see the errors in the macro expansion window (and at the instruction itself), but it is probably too little time to do that right now..

PR #290, opened 2017-01-26
Feature/editor context menu

goldsborough:
Refactors the Editor and gives it a context menu


andizimmerer (commented):
the text in the first line of the editor is not visible. there is something that overlays this line (and only the first line).

line_overlay

both lines contain `gagagag` (the lower part of the `g` is still visible in the first line.

--edit---
Everything works fine after updating to the latest theme.

Jibbow:
both lines contain `gagagaga` (the lower part of the `g` is visible.

FelixOpolka:
Binary formatting in registers is broken. It works on master, though.

FelixOpolka (approved):
When pressing "Copy All", the cursor is moved to end of file. Maybe you can restore its previous position after copying. Otherwise everything looks good!

PR #293, opened 2017-01-27
Feature/theme input

### LINKED ISSUE #233: Enhance console/merge input and ouput console
See comments here: https://github.com/TUM-LRR/era-gp-sim/pull/214/files

janschopohl:
closes #233 #283 #292


riedersa:
In my opinion it's fine and it looks greate! But maybe the Icon for the input could be changed?

riedersa (approved):

source/ui/Components/Output/Console/ConsoleSettingsWindow.qml
```diff
@@ -1,159 +1,169 @@
...
-        baseAddressTextField.text = outputComponent.getOutputItem(outputItemIndex)["baseAddress"];
-        var t = outputComponent.getOutputItem(outputItemIndex)["textMode"];
-        if(t === 0){
-            ab.checked = true;
+  id: settingsWindowConsole
+  width: Theme.console.settings.width
+  height: Theme.console.settings.height
+  property int outputItemIndex: 2
+
+  title: "Console Settings"
```
andizimmerer: Having a short explanation for array mode and pipe mode in the settings window would be nice :)
But this is totally optional.

source/ui/console-component.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+void ConsoleComponent::resetInterrupt() {
+  MemoryValue memoryValue =
+      _memoryAccess.getMemoryValueAt(_interruptAddress).get();
+  uint8_t value = conversions::convert<uint8_t>(memoryValue);
+  MemoryValue resetValue = conversions::convert(value & 254, 8);
+  _memoryAccess.putMemoryValueAt(_interruptAddress, resetValue);
+}
+
+void ConsoleComponent::deleteTextInMemory() {
```
andizimmerer, line 141: You should do a reparse here. Consider the following scenario:
the user reserves the first few memory cells with `.resb`. When you clear the console now the reserved bytes will be set to 0. A following execution of the program could now function incorrectly because the memory contains unexpected values.

themes/default.theme/input.sass
```diff
@@ -4,3 +4,42 @@ input
...
+      color: "black"
+      background: $divider-color
+      animation
+        marginAnimation
+          duration: 300
+        colorAnimation
+          duration: 600
+    active
+      margins: 20
+      color: "black"
```
andizimmerer, line 15: I think white would look better ;)
And definitely use a larger font size otherwise the text looks so lost

andizimmerer (commented):
Beautiful! :)

andizimmerer (approved):

source/ui/console-component.cpp
janschopohl: Isn't that the point of the console though? Otherwise you could make the argument that you should reparse after every change in memory...
I think console input and clearing is something where the user has to expect changing the memory, as described in the user manual.

themes/default.theme/input.sass
janschopohl: Someone else actually requested black :smile:

But that's the point of themes! We need more anyways, if you have time to spare, you could start a new theme!

source/ui/Components/Output/Console/ConsoleSettingsWindow.qml
janschopohl: I actually considered this, but I couldn't come up with a precise, yet short (enough) explanation. As we write a user manual, I think this is ok.
If you have concrete suggestions, feel free to add them/send them to me.

source/ui/console-component.cpp
andizimmerer: I just wanted to mention that because I stumbled across it when i was trying to set a "default text" to the console with `.resb`. When I cleared the console the default text was lost during the next execution

PR #294, opened 2017-01-27
Feature/better syntax highlighting

kreudom:
Instead of highlighting only known instructions, all words at the start of the line or following a `:` are now highlighted as instructions. This allows macros to be highlighted without making too many changes to the rest of the code.


source/parser/riscv/riscv-parser.cpp
```diff
@@ -225,11 +214,10 @@ const SyntaxInformation RiscvParser::getSyntaxInformation() {
...
 
   // Add immediate regex
-  // Matches arithmetic expressions containing digits, operators, brackets and
-  // spaces. Expressions need to start with a digit, an open bracket or an unary
-  // operator.
+  // Matches expressions containing digits, letters, operators, brackets and
+  // spaces.
   info.addSyntaxRegex(
-      R"(\b[\+\-0-9\(!~][0-9a-fA-Fx\+\-%\*\/\(\)\|\^&=!<>~\t ]*)",
+      R"([0-9a-zA-Z\+\-%\*\/\(\)\|\^&=!<>~_]+)",
```
davschneller: Maybe making labels/constants (or all "words") highlighted differently in here? (no required, but it would be nice in my opinion)
Also, maybe the constant numbers in a different color? Or is this too late to change?

source/parser/riscv/riscv-parser.cpp
```diff
@@ -200,20 +200,9 @@ FinalRepresentation RiscvParser::parse(const std::string& text) {
...
-  // Add directive regexes
-  for (auto directive : RiscVDirectiveFactory::mapping) {
-    // Matches all directive mnemonics starting with a '.' which don't end with
-    // a ':'
-    info.addSyntaxRegex("\\." + directive.first + "\\b(?!:)",
-                        SyntaxInformation::Token::Instruction);
-  }
+  // Matches single words after a : or line-start
+  info.addSyntaxRegex(R"((?<=^|:) *[a-zA-Z.]+)",
+                      SyntaxInformation::Token::Instruction);
```
davschneller, line 212: Generally, shouldn't enum names be all uppercase?

source/parser/riscv/riscv-parser.cpp
```diff
@@ -225,11 +214,10 @@ const SyntaxInformation RiscvParser::getSyntaxInformation() {
   info.addSyntaxRegex("^\\s*\\w+:", SyntaxInformation::Token::Label);
```
davschneller: Might this not be able to be applied rather to all words? (including constants) Or what do you think of this?

davschneller (approved):
I like it!

goldsborough (approved):

ErikP0 (approved):
Works as intended

PR #297, opened 2017-01-28
Memory is resized if a snapshot is bigger than the current size.

janschopohl:
Memory size can only increase, not decrease.

### DIFF
--- a/include/core/memory.hpp
+++ b/include/core/memory.hpp
@@ -31,12 +31,13 @@
 
 class Memory {
  public:
+  using size_t = std::size_t;
   using Json = nlohmann::json;
-  using CallbackFunction =
-      std::function<void(const std::size_t, const std::size_t)>;
-  using RawMapPair = std::pair<std::map<std::string, std::size_t>,
+  using UpdateCallback = std::function<void(const size_t, const size_t)>;
+  using SizeCallback = std::function<void(const size_t)>;
+  using RawMapPair = std::pair<std::map<std::string, size_t>,
                                std::map<std::string, std::string>>;
-  using ProtectionMap = std::map<std::size_t, std::size_t>;
+  using ProtectionMap = std::map<size_t, size_t>;
   /**
    * \brief Default constructor. Constructs an empty Memory with default size
    *        (64 Bytes � 8 Bit)
@@ -48,7 +49,7 @@ class Memory {
    * \param byteCount Number of Bytes
    * \param byteSize Size of a Byte in Bit
    */
-  Memory(std::size_t byteCount, std::size_t byteSize = 8);
+  Memory(size_t byteCount, size_t byteSize = 8);
 
   /**
    * \brief Constructs an Memory with the data stored in json
@@ -93,19 +94,24 @@ class Memory {
    * \brief returns _byteSize
    * \returns returns _byteSize
    */
-  std::size_t getByteSize() const;
+  size_t getByteSize() const;
 
   /**
   * \brief returns _byteCount
   * \returns returns _byteCount
   */
-  std::size_t getByteCount() const;
+  size_t getByteCount() const;
 
   /**
    * \brief Sets the callback to notify the gui about changes in the data
    * \param callback the callback to be set as _callback
    */
-  void setCallback(const CallbackFunction &callback);
+  void setCallback(const UpdateCallback &callback);
+
+  /**
+   * Sets the callback which notifies the gui about a change of the memory size.
+   */
+  void setSizeCallback(const SizeCallback &callback);
 
   /**
    * \brief Returns a MemoryValue holding the data stored in the Memory at
@@ -117,7 +123,7 @@ class Memory {
    * \throws AssertionError iff [address; address+amount[ is not subset of
    *         [0; _byteCount[
    */
-  MemoryValue get(std::size_t address, std::size_t amount = 1) const;
+  MemoryValue get(size_t address, size_t amount = 1) const;
   /**
    * \brief Returns a MemoryValue holding the data stored in the Memory at
    *        [address; address+amount[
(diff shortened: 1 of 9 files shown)


include/core/memory.hpp
```diff
@@ -337,40 +340,44 @@ class Memory {
...
+  /** This callback is called when the memory size changes. */
+  SizeCallback _sizeCallback;
 
   /**
    * \brief vector storing data about protected memory areas
-   * \note this takes linear time, one could improve this by implementing some
+   * \note this takes linear time, one could improve this by implementing
+   * some
    *       binary search, or using some indexed algorithm
    */
```
ProfPille, line 378: format?

source/core/memory.cpp
```diff
@@ -291,36 +292,33 @@ void Memory::deserializeJSON(const Json& json) {
...
-    std::string recieved = std::to_string(byteCount);
-    throw DeserializationError(
-        "Could not deserialize Memory: ByteCount did not match: expected \"" +
-        expected + "\" to be equal to \"" + recieved + "\"");
+  size_t byteCount = *byteCountIt;
+  if (_byteCount < byteCount) {
+    _byteCount = byteCount;
+    _data = MemoryValue(byteCount * _byteSize);
+    _sizeCallback(_byteCount);
   }
```
ProfPille, line 308: what about smaller memories?

source/core/memory.cpp
```diff
@@ -291,36 +292,33 @@ void Memory::deserializeJSON(const Json& json) {
...
         expected + "\" to be equal to \"" + recieved + "\"");
   }
-  std::size_t lineLength = json[_lineLengthStringIdentifier].get<std::size_t>();
+  size_t lineLength = json[_lineLengthStringIdentifier].get<size_t>();
   if (lineLength <= 0) {
     throw DeserializationError("Could not deserialize Memory: lineLength == 0");
   }
 
   // clear the current memory to write the zeros in the serialized data
   clear();
```
ProfPille: You could optimize the usage of clear and avoid clearing if you resize the memory

source/core/memory.cpp
```diff
@@ -291,36 +292,33 @@ void Memory::deserializeJSON(const Json& json) {
...
-    std::size_t byteCount = *byteCountIt;
-    std::string recieved = std::to_string(byteCount);
-    throw DeserializationError(
-        "Could not deserialize Memory: ByteCount did not match: expected \"" +
-        expected + "\" to be equal to \"" + recieved + "\"");
+  size_t byteCount = *byteCountIt;
+  if (_byteCount < byteCount) {
+    _byteCount = byteCount;
+    _data = MemoryValue(byteCount * _byteSize);
+    _sizeCallback(_byteCount);
```
ProfPille, line 303: it'd be also nice if you made a private method that calls the sizeCallback, just like the updateCallback, as that makes this more easily expandable

source/core/memory.cpp
```diff
@@ -37,11 +37,11 @@ const std::string Memory::_dataMapStringIdentifier = "memory_map";
 Memory::Memory() : Memory(64, 8) {
 }
 
-Memory::Memory(std::size_t byteCount, std::size_t byteSize)
+Memory::Memory(size_t byteCount, size_t byteSize)
 : _byteCount{byteCount}
 , _byteSize{byteSize}
 , _data{byteCount * byteSize}
-, _callback{[](const std::size_t, const std::size_t) {}} {
+, _callback{[](const size_t, const size_t) {}} {
```
ProfPille: I don't think these need to be `const size_t`

include/core/memory.hpp
```diff
@@ -328,7 +331,7 @@ class Memory {
    * \param address address of the updated value
    * \param amount length of the updated value
    */
-  void _wasUpdated(std::size_t address, std::size_t amount = 1) const;
+  void _wasUpdated(size_t address, size_t amount = 1) const;
```
ProfPille, line 334: Like this

include/core/memory.hpp
```diff
@@ -31,12 +31,13 @@
 
 class Memory {
  public:
+  using size_t = std::size_t;
   using Json = nlohmann::json;
-  using CallbackFunction =
-      std::function<void(const std::size_t, const std::size_t)>;
-  using RawMapPair = std::pair<std::map<std::string, std::size_t>,
+  using UpdateCallback = std::function<void(const size_t, const size_t)>;
+  using SizeCallback = std::function<void(const size_t)>;
```
ProfPille: I don't think we need `const size_t` here

ProfPille (changes requested):

source/core/memory.cpp
janschopohl: Reducing the size of the memory can cause problems with gui components, which might not adapt to the new memory size immediately. This is not critical when increasing the size, but can cause crashes when decreasing the size.
Because of that, the size is only changed if the snapshot would not fit into the current memory.

It's also part of the description of this pull request...

source/core/memory.cpp
janschopohl: I don't really know what you want ot expand there. Adding another private method just adds redundancy, in my opinion...

include/core/memory.hpp
janschopohl: Seems like it wasn't formatted properly before and got reformatted automatically on save.

source/core/memory.cpp
ProfPille: you could add a comment saying that, otherwise someone might accidentally try to implement exactly that

source/core/memory.cpp
ProfPille: maybe someday someone decides to add multible subscribers, or keeps track of the calls for Example

source/core/memory.cpp
```diff
@@ -291,13 +291,19 @@ void Memory::deserializeJSON(const Json& json) {
                                _dataMapStringIdentifier);
   }
   // TODO::check types!
-  // check that sizes match
+  // check wether the sizes match
```
ProfPille: "whether"
"wether" means "Hammel" (kastrierter Ziegenbock)

ProfPille (changes requested):

ProfPille (approved):

PR #303, opened 2017-01-29
Hotfix/tabbar-close

### LINKED ISSUE #302: Tabbar close buttons always close current project

FelixOpolka:
Closes #302
Fix bug that would cause wrong project to close when pressing tab’s
close button

### DIFF
--- a/source/ui/Components/Menubar/Menubar.qml
+++ b/source/ui/Components/Menubar/Menubar.qml
@@ -88,7 +88,7 @@ MenuBar {
     MenuItem {
       text: "Close Project"
       shortcut: "Ctrl+W"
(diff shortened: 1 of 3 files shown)


riedersa:
When starting the program i get the following message:
qrc:/Components/ProjectCreation/Sections/ProjectName.qml:41: TypeError: Cannot read property 'fontSize' of undefined - was a theme errror, now it works fine
And when closing a creation screen while cursor is in textfield, i get:
file:///home/sabine/Qt/5.6/gcc_64/qml/QtQuick/Controls/TextField.qml:667: TypeError: Cannot call method 'editingFinished' of null

PR #326, opened 2017-02-02
Populate .erasim directory on startup

k-lukas:
Creates the basic structure of the .erasim config folder automatically. It is assumed, that the final application executable is in the same folder as the `themes/` directory, so the default theme can be copied.
(If that's not the case, create a symlink if necessary `ln -s /path/to/project/themes ./themes`)

### DIFF
--- a/include/ui/theme.hpp
+++ b/include/ui/theme.hpp
@@ -21,6 +21,7 @@
 #define ERAGPSIM_UI_THEME_HPP
 
 #include <QByteArray>
+#include <QDir>
 #include <QHash>
 #include <QJsonObject>
 #include <QQmlPropertyMap>
@@ -115,6 +116,17 @@ class Theme : public QQmlPropertyMap {
 
   /** The singleton. */
   static Theme* _theme;
+
+  /**
+   * Attempts to copy a theme from the resource directory (located in the
+   * theme/ directory relative to the executable/working directory) into the
+   * given destination directory.
+   *
+   * \param name The name of the theme to copy.
+   * \param destination the destination directory, to copy the theme into.
+   * \return true if theme could be copied into the destination directory.
+   */
+  static bool _copyTheme(const QString& name, QDir destination);
 
   /**
    * Loads raw data from disk for a given theme.
--- a/source/ui/settings.cpp
+++ b/source/ui/settings.cpp
@@ -134,7 +134,7 @@ Settings::Json Settings::toJson() const {
 }
 
 StatusWithValue<QString> Settings::_findSettingsDirectory() {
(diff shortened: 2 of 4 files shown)


source/ui/settings.cpp
```diff
@@ -134,13 +134,18 @@ Settings::Json Settings::toJson() const {
 }
 
 StatusWithValue<QString> Settings::_findSettingsDirectory() {
+  const auto path = ".erasim";
```
goldsborough: `static`

source/ui/theme.cpp
```diff
@@ -84,17 +84,55 @@ const QString& Theme::currentThemeName() const noexcept {
   return _currentThemeName;
 }
 
+bool Theme::_copyTheme(const QString& name, QDir destination) {
+  // Obtain the executable directory
+  QDir directory{QCoreApplication::applicationDirPath()};
```
goldsborough: Please use `=` or parantheses `(...)`

source/ui/theme.cpp
```diff
@@ -84,17 +84,55 @@ const QString& Theme::currentThemeName() const noexcept {
...
+    return false;
+  }
+
+  if (!destination.cd(name + ".theme")) {
+    return false;
+  }
+
+  // Copy theme files
+  for (const auto& fileInfo : directory.entryInfoList(QDir::Files)) {
+    QFile file{fileInfo.absoluteFilePath()};
```
goldsborough: `=` or `(...)`

include/ui/theme.hpp
```diff
@@ -117,6 +118,17 @@ class Theme : public QQmlPropertyMap {
...
   /**
+   * Attempts to copy a theme from the resource directory (located in the
+   * theme/ directory relative to the executable/working directory) into the
+   * given destination directory.
+   *
+   * \param name The name of the theme to copy.
+   * \param destination the destination directory, to copy the theme into.
+   * \return true if theme could be copied into the destination directory.
+   */
+  static bool _copyTheme(const QString& name, QDir destination);
```
goldsborough, line 129: `const QDir&` please

goldsborough (approved):
LGTM

include/ui/theme.hpp
k-lukas: Actually, there is a reason for that: destination is changed during `_copyTheme`, due to a call to `cd()`. If destination it's a const reference, `cd()` can't be called and if its just a reference, the calling function won't work correctly. So i figured, call by value would be the best in this case.

goldsborough: ok

PR #330, opened 2017-02-04
symbol for write protected memory cells

andizimmerer:


k-lukas (approved):

ErikP0:
Uknown Characters This happens on Windows with Qt 5.8 and on my Ubuntu VM with (broken) Qt 5.6.2.
Can you reproduce this, or am I missing a font?

janschopohl:
In the default theme, for me it looks like this:
grafik

However, there is a different symbol for the presentation theme. Is this on purpose?
grafik

The alignement is also not quite perfect :wink:

andizimmerer:
@ErikP0 good point! Is fixed now.
@janschopohl yes, was intended, but now I changed both to the same symbol (and I've changed the alignment a little) ;)

ErikP0 (approved):
Ok after manually updating the them, it works!

PR #331, opened 2017-02-04
Installer

k-lukas:
This pull request adds the installer for Windows & Linux.

Furthermore, the following notable changes were made:
* Add Qt5::Svg as dependency for the gui (needed for rendering SVG icons)
* Use CMake build types. There are 2 major types: `Debug` and `Release`. The first one resembles the previous setup, so it adds debug symbols and is the default build type. The second one is used for releasing the software and doesn't include debug symbols, which greatly reduces the size of the final executable. But notice, that some tests won't work with the Release build type.
* If a Qt instance is running, resources (`isa/` and `themes`) are now loaded relative to the executable location (have a look at the changes made to `utility.cpp`). On Unix, symlinks are created in the build directory, so you can use it as before.


andizimmerer (approved):

goldsborough:
Fix merge conflicts

goldsborough (approved):

PR #332, opened 2017-02-05
Build fixes for Visual Studio

engelmarkus:
I've fixed a few things in the source to be able to build on Windows using Visual Studio 2017 (older versions won't work because the necessary <optional> header is missing).

Some of these commits are simply bug fixes that should be merged anyway :) .

### DIFF
--- a/source/parser/riscv/riscv-directive-factory.cpp
+++ b/source/parser/riscv/riscv-directive-factory.cpp
@@ -51,15 +51,14 @@ template <typename T>
 const ExpressionCompiler<T>
     _expressionCompiler = CLikeExpressionCompilers::createCLikeCompiler<T>();
 
-// Function for `createMemoryDefinitionDirective`
+
+namespace {
 template <typename T>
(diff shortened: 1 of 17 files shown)


source/parser/riscv/riscv-directive-factory.cpp
```diff
@@ -51,15 +51,14 @@ template <typename T>
...
+namespace {
 template <typename T>
-const typename MemoryDefinitionDirective<T>::ProcessValuesFunction
-    _processMemoryDefinitionValues =
-        [](const PositionedStringVector &values,
-           const SymbolReplacer &replacer,
-           std::size_t cellSize,
-           CompileErrorList &errors,
-           const std::function<void(T, std::size_t)> &handler) -> std::size_t {
+std::size_t _processMemoryDefinitionValuesLambda(const PositionedStringVector &values,
```
goldsborough: \>80 zeichen du verbrecher
lauf bitte da mal mit unserem `.clang-format` (im repo) mit clang-format drueber, dann passt

goldsborough (approved):

Issue #355, opened 2017-03-12
editor: moving the cursor activates menu


meyerm:
When moving the cursor above the first/under the last line in the editor view, the menu gets activated.

PR #368, opened 2017-04-27
Should fix #346, also prevents #355

### LINKED ISSUE #345: editor: error messages overlap the command
Error messages aren't wrapped into the next line early enough. F.ex. writing `addi x1, x2` and resizing the editor frame shows that the error message overlaps with `, x2` before it is finally shifted one line lower.

### LINKED ISSUE #346: editor: text view doesn't follow the cursor
Navigating in the assembler code won't slide the view on the source within the editor. You have to scroll with the mouse to see where your cursor is.

janschopohl:
closes #346 #355

### DIFF
--- a/source/ui/Components/editor/Editor.qml
+++ b/source/ui/Components/editor/Editor.qml
@@ -76,6 +76,8 @@ ScrollView {
       id: item
       anchors.fill: parent
 
+      property alias zoom: scale.zoom
+
       // Scaling transofmation for zooming
       transform: Scale {
         id: scale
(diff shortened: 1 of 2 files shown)


goldsborough (approved):

OlafLostViking:
I tried it out and it seems to fix #345 :-). Unfortunately, #355 is still valid.

Thank you very much for having a look!

meyerm:
Sry, added my comment to the merge... :1st_place_medal:

I tried it out and it seems to fix #345 :-). Unfortunately, it looks like #355 is still valid (at least on my system).
Thank you very much for having a look!

janschopohl:
Oops, I somehow mistyped and wrote 345 instead of 346. This only fixes #346.

